PR #5, opened 2019-11-20
Initial Serilog JSON formatter for discussion

codebrain:
This is a draft PR aimed to open discussion around a Serilog ECS integration.

This enriches and essentially re-maps properties on the `logEvent` argument and surrounding environment into the ECS format in a JSON format (using UTF8JSON), which can then be redirected to a sink of your own choice.

Additional properties can be set in the `appSettings` keys for inclusion in the ECS payload.

```
    <appSettings>
        <add key="ApplicationId" value="MyApp" />
        <add key="ApplicationName" value="My Application" />
        <add key="ApplicationType" value="API" />
        <add key="ApplicationVersion" value="1.0.0" />
    </appSettings>
```

An example of the rendered JSON from the `Elastic.CommonSchema.Serilog.Test` project:

```
{
  "ecs": {
    "version": "1.2.0"
  },
  "event": {
    "severity": 0,
    "timezone": "AUS Eastern Standard Time",
    "created": "2019-11-22T14:59:02.5903135+11:00"
  },
  "log": {
    "level": "Information",
    "logger": "Elastic.CommonSchema.Serilog"
  },
  "process": {
    "thread": {
      "id": 1
    },
    "executable": "System.Threading.ExecutionContext"
  },
  "@timestamp": "2019-11-22T14:59:02.5903135+11:00"
}
```

PR #6, opened 2019-11-27
Feature/serialization changes

Mpdreamz:
Opening up as a PR to the #5 PR.

The serialization routines should live in `Elastic.CommonSchema` so that all integrations can take advantage of it. 

This introduces dedicated formatters for `Base` and `Log` to emit the same way https://github.com/elastic/ecs-logging-java

Does.


src/Elastic.CommonSchema/Location.cs
```diff
@@ -1,3 +1,20 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
```
Mpdreamz: We can use the condensed header here (3 lines)

++ For starting including though

src/Generator/Views/BaseJsonFormatter.Generated.cshtml
```diff
@@ -0,0 +1,171 @@
...
+@foreach (var field in Model.BaseFieldsOrdered)
+{
+    if (field.Item1 != "WriteProp")
+    {<text>            @(field.Item1)(ref writer, value, formatterResolver); // @(index++)
+</text>}
+    else
+    {<text>            @(field.Item1)(ref writer, "@field.Item2.Name", value.@FileGenerator.PascalCase(field.Item2.Name), formatterResolver); // @(index++)
+</text>}
+}
+            // Complex types
```
Mpdreamz: Love this but since this is the only bit that needs generation perhaps we can rely on this class being partial and we generate a method stub we can call from the manually implemented side of this formatter?

src/Generator/Views/Types.Generated.cshtml
```diff
@@ -164,7 +171,7 @@ namespace Elastic.CommonSchema
     /// </remarks>
 </text>
     }
-    public class @FileGenerator.PascalCase(schema.Name) 
+    public @(schema.Name == "log" ? "partial " : "")class @FileGenerator.PascalCase(schema.Name)
```
Mpdreamz: Should we just generate all as partial? Allows us to hook in in the future if we need to?

src/Elastic.CommonSchema/Serialization/LogJsonFormatter .cs
```diff
@@ -1,12 +1,28 @@
...
 using System;
-using System.Collections.Generic;
 using Utf8Json;
 
 namespace Elastic.CommonSchema.Serialization
 {
     internal class LogJsonFormatter : IJsonFormatter<Log>
     {
-        private static readonly AutoAutomataDictionary AutomataDictionary = new AutoAutomataDictionary
+        private static readonly IncrementingAutomataDictionary AutomataDictionary = new IncrementingAutomataDictionary
```
Mpdreamz: Much saner name :)

src/Generator/Views/Types.Generated.cshtml
codebrain: Yup, makes sense, will follow up with another commit.

PR #7, opened 2019-11-28
move to ECS 1.3

Mpdreamz:


codebrain:
Looks good to me, the only thing is I had to manually add some special cases into the generator so that the properties are rendered as `string[]`.

https://github.com/elastic/ecs-dotnet/pull/7/commits/ff92720b40748e76887cb91e810e161af787559f

The specification currently doesn't have an attribute to indicate the property type is an array, but there is somewhat of a pattern where the description will start with "Array of...".

PR #8, opened 2019-12-02
Tests

Mpdreamz:
This adds a bunch of tests

* make sure `Base` can roundtrip serialization
* Add infra to easily test serilog integration
* Make sure we play nicely with default enrichers
* Prevent properties already persisted in the ECS tree to be written again under meta data
* metadata keys and object keys are now snake_case by default as per ECS recommendations.
   - This does not generically apply to all Dictionaries though
* make sure message is persisted

And various other fixes

### DIFF
--- a/src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
+++ b/src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Threading;
 using Serilog.Events;
 #if NETSTANDARD
@@ -23,6 +24,28 @@ namespace Elastic.CommonSchema.Serilog
 	/// </summary>
 	public class LogEventConverter
 	{
+		private static class SpecialKeys
+		{
+			public const string DefaultLogger = "Elastic.CommonSchema.Serilog";
+
+			public const string SourceContext = nameof(SourceContext);
+			public const string EnvironmentUserName = nameof(EnvironmentUserName);
+			public const string Host = nameof(Host);
+			public const string ActionCategory = nameof(ActionCategory);
+			public const string ActionName = nameof(ActionName);
+			public const string ActionId = nameof(ActionId);
+			public const string ActionKind = nameof(ActionKind);
+			public const string ActionSeverity = nameof(ActionSeverity);
+			public const string ApplicationId = nameof(ApplicationId);
+			public const string ApplicationName = nameof(ApplicationName);
+			public const string ApplicationType = nameof(ApplicationType);
+			public const string ApplicationVersion = nameof(ApplicationVersion);
+			public const string ProcessName = nameof(ProcessName);
+			public const string ProcessId = nameof(ProcessId);
+			public const string ThreadId = nameof(ThreadId);
+			public const string MachineName = nameof(MachineName);
+		}
+
 		public static Base ConvertToEcs(LogEvent logEvent, IEcsTextFormatterConfiguration configuration)
 		{
 			var exceptions = logEvent.Exception != null
@@ -38,18 +61,10 @@ namespace Elastic.CommonSchema.Serilog
 				Log = GetLog(logEvent, exceptions, configuration),
 				Agent = GetAgent(logEvent),
 				Event = GetEvent(logEvent),
-				Metadata = GetMetadata(logEvent)
+				Metadata = GetMetadata(logEvent),
+				Process = GetProcess(logEvent, configuration.MapCurrentThread),
+				Host = GetHost(logEvent),
 			};
-
-			//TODO investigate
-			//Serilog sinks with default enrichments where do these end up?
-			//logEvent.Properties
-
-			if (configuration.MapCurrentThread)
-			{
-				var currentThread = Thread.CurrentThread;
-				ecsEvent.Process = GetProcess(currentThread);
-			}
 
 			if (configuration.MapHttpAdapter != null)
 			{
@@ -68,10 +83,11 @@ namespace Elastic.CommonSchema.Serilog
 			return ecsEvent;
 		}
 
+
 		private static IDictionary<string, object> GetMetadata(LogEvent logEvent)
 		{
 			var dict = new Dictionary<string, object>();
-
+			//TODO what does this do and where does it come from?
 			if (logEvent.Properties.TryGetValue("ActionPayload", out var actionPayload))
 			{
 				var logEventPropertyValues = (actionPayload as SequenceValue)?.Elements;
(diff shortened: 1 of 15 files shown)


src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
```diff
@@ -85,55 +103,167 @@ public static Base ConvertToEcs(LogEvent logEvent, IEcsTextFormatterConfiguratio
...
+				else
+					dict.Add(ToSnakeCase(logEventPropertyValue.Key), logEventPropertyValue.Value);
 			}
-
+			if (dict.Count == 0) return null;
 			return dict;
 		}
 
+		//TODO this should live in Log.MetaData as custom dictionary converter
+		private static string ToSnakeCase(string s)
```
codebrain: ++ to this, but I also wonder if we ought to also store the original value before it was transformed?

`ValueX` -> `value_x`

But perhaps also store `original: "ValueX"` or something similar?

src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
```diff
@@ -85,55 +103,167 @@ public static Base ConvertToEcs(LogEvent logEvent, IEcsTextFormatterConfiguratio
...
 
-			var hasHost = e.Properties.TryGetValue("Host", out var host);
+			var hasHost = e.Properties.TryGetValue(SpecialKeys.Host, out var host);
 			server.Address = hasHost ? host.ToString() : null;
 			server.Ip = hasHost ? host.ToString() : null;
 			return server;
 		}
 
-		private static Process GetProcess(Thread currentThread)
+		private static Process GetProcess(LogEvent e, bool mapFromCurrentThread)
```
codebrain: This is way nicer, using the enricher property keys

src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
```diff
@@ -85,55 +103,167 @@ public static Base ConvertToEcs(LogEvent logEvent, IEcsTextFormatterConfiguratio
...
+		}
+
 		private static Log GetLog(LogEvent e, IReadOnlyList<Exception> exceptions, IEcsTextFormatterConfiguration configuration)
 		{
-			var log = new Log { Level = e.Level.ToString("F"), Logger = "Elastic.CommonSchema.Serilog" };
+			var source = e.Properties.TryGetValue(SpecialKeys.SourceContext, out var context)
+				 ? context.ToString()
+				 : SpecialKeys.DefaultLogger ;
+
+			var log = new Log { Level = e.Level.ToString("F"), Logger = source};
```
codebrain: IIRC I believe we are also rendering out the numeric equivalent in the json?

We'd really want to run queries where log.level >= 1 (for example)

src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
```diff
@@ -153,53 +283,45 @@ private static Event GetEvent(LogEvent e)
...
-					? long.Parse(e.Properties["ActionSeverity"].ToString())
-					: 0
+				Timezone = TimeZoneInfo.Local.StandardName
 			};
 
-#if NETSTANDARD
-            evnt.Timezone = TimeZoneInfo.Local.StandardName;
-#else
+			//Why does this get overriden in full framework?
+#if FULLFRAMEWORK
```
codebrain: Good question, it was in there as part of an `if` back when it was switching on `DOTNETCORE` - can now be deleted due to line 297 above.

src/Elastic.CommonSchema/Serialization/BaseJsonFormatter.Generated.cs
```diff
@@ -86,13 +87,13 @@ static bool ReadString(ref JsonReader reader, ref string stringProp)
...
 			}
-			
+
 			while (reader.ReadIsInObject(ref count))
 			{
 				var propertyName = reader.ReadPropertyNameSegmentRaw();
 				if (AutomataDictionary.TryGetValue(propertyName, out var value))
 				{
-					_ = value switch
+					var read = _ = value switch
```
codebrain: Dont forget to update the .cshtml template and regen.

src/Elastic.CommonSchema/Serialization/LogJsonFormatter .cs
```diff
@@ -3,13 +3,79 @@
...
+		}
+	}
+
+
+	internal class LogJsonFormatter : EcsUtf8JsonFormatterBase<Log>, IJsonFormatter<Log>
+	{
+		private static IncrementingAutomataDictionary AutomataDictionary { get; } = new IncrementingAutomataDictionary
+		{
+			// Base fields
+			{ "origin" }, // 0
```
codebrain: Will follow up in a PR next week to generate this too... we dont want to miss fields.

src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
Mpdreamz: Cool cool, we'll delete it at some point :)

src/Elastic.CommonSchema/Serialization/BaseJsonFormatter.Generated.cs
Mpdreamz: Updated as part of the move to `System.Text.Json` https://github.com/elastic/ecs-dotnet/pull/9

src/Elastic.CommonSchema/Serialization/LogJsonFormatter .cs
Mpdreamz: ++

src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
Mpdreamz: This is now handled in the json converters in https://github.com/elastic/ecs-dotnet/pull/9

Since we only snake_case keys since all of  ECS is snake_case I don't think there is usecase to introduce the original key value.

Do you have a usecase in mind?

src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
Mpdreamz: Ahh we are only emitting the `log.level` as string, mind need an additional field in ECS?

src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
codebrain: The only reason was for if an external system (or user) wanted to correlate back to the original token name in the source code. I was thinking if we stored the original parameter name that would make it possible without having to UnSnakeCase (and potentially handle any mismatches)

src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
codebrain: Its a very good question, I raised as an issue in the repo - https://github.com/elastic/ecs/issues/683

codebrain: Opened: https://github.com/elastic/ecs-dotnet/pull/11

PR #9, opened 2019-12-03
Feature/system.text.json

Mpdreamz:
* Move to `System.Text.Json`
* Add more overloads for Serialize/Deserialize (Stream)
* Async support
* Fast path for `StreamWriter` for Serilog integration


src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
```diff
@@ -144,33 +145,7 @@ public static Base ConvertToEcs(LogEvent logEvent, IEcsTextFormatterConfiguratio
...
-					sb.Append(char.ToLowerInvariant(c));
-				else
-				{
-					sb.Append("_");
-					sb.Append(char.ToLowerInvariant(c));
-				}
-			}
-			return sb.ToString();
-		}
+		private static string ToSnakeCase(string key) => key;
```
Mpdreamz: Can be removed

src/Elastic.CommonSchema/Serialization/BaseJsonFormatter.Generated.cs
```diff
@@ -5,233 +5,124 @@
...
-					}
-				}
-				else reader.ReadNext();
-			}
-			ecsEvent.Log ??= new Log();
-			ecsEvent.Log.Level = loglevel;
-			ecsEvent.Message = message;
-			ecsEvent.Timestamp = timestamp;
-			return ecsEvent;
+				"@timestamp" => ReadDateTime(ref reader, ref @timestamp),
```
gregkalapos: I see this file is autogenerated, but wow the method parameter ended up being `ref @timestamp` - which as it seems does not change the variable name 😅  - so `@timestamp` is the same as `timestamp`. Very funky.

src/Elastic.CommonSchema/Serialization/BaseJsonConverter.cs
```diff
@@ -0,0 +1,51 @@
...
+			DateTimeOffset? timestamp = default;
+			while (reader.Read())
+			{
+				if (reader.TokenType == JsonTokenType.EndObject)
+					break;
+
+				if (reader.TokenType != JsonTokenType.PropertyName)
+					throw new JsonException();
+
+				var read = ReadProperties(ref reader, ecsEvent, ref timestamp, ref loglevel);
```
gregkalapos: nit: `read` is ignored - not sure there could be anything done here if it's `false`.

src/Elastic.CommonSchema/Serialization/BaseJsonFormatter.Generated.cs
Mpdreamz: Hehe @ allows you to use otherwise reserved keywords as variablenames but can be used in front of all variable names however nonsensical :)

src/Elastic.CommonSchema/Serialization/BaseJsonConverter.cs
Mpdreamz: Aye still undecided what to here, we could ignore, throw, read unknown kv pairsand store on metadata. Will open a follow-up ticket.

Mpdreamz:
> Now I feel really motivated to add the trace.id and transaction.id parsing part to put those into ecs with the new enricher rocket

Please do :)

src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
Mpdreamz: Will follow up

PR #10, opened 2019-12-03
Add Serilog enricher to append APM transaction and trace id to logs

gregkalapos:
This PR adds an enricher to correlate Serilog logs and traces from Elastic APM. 

```
var logger = new LoggerConfiguration().Enrich.WithElasticApmTraceId()
```

With the above line each time a transaction is created by the APM Agent, the id of the transaction and the trace id is added to all logs during the transaction as properties.

Which is used in Kibana here:

Screen Shot 2019-12-03 at 22 10 18

Naming is totally up for discussion. Also, I just added the projects to `/src` - happy to move it if we feel we should.


gregkalapos:
- Moved test to the `tests` folder. 
- Changed the name according to https://github.com/elastic/apm-agent-dotnet/pull/634#discussion_r353585082 
- Added a readme 

I can't merge, because I don't have write access to this repo 👼

Mpdreamz:
Try now :)

PR #12, opened 2019-12-10
Read trace and transaction id from APM in EcsTextFormatter

gregkalapos:
Adds trace and transaction id to the generated logs in `EcsTextFormatter` in case the APM Agent is active.

Configuration:

```
Log.Logger = new LoggerConfiguration()
.Enrich.WithElasticApmCorrelationInfo()
.WriteTo.Elasticsearch(new ElasticsearchSinkOptions(new Uri("http://localhost:9200"))
{
  CustomFormatter = new EcsTextFormatter()
})
.CreateLogger();
```

On this screenshot you see the output when I pass `EcsTextFormatter` to the serilog elasticsearch sink.

Screen Shot 2019-12-10 at 17 20 27

I ran into 1 problem during testing with ASP.NET Core: It seems that some component uses the property `traceId` (my suspect is [serilog-aspnetcore](https://github.com/serilog/serilog-aspnetcore)), so our trace Id was overwritten by the Activity Id (`Activity.Current.Id`) and broke the navigation between logs and traces in Kibana.

This is yet another reason we should reuse that id. Nevertheless we don't have it right now, so I thought we just prefix our properties with `ElasticApm`, so we store the traceId with `logEvent.AddPropertyIfAbsent(propertyFactory.CreateProperty("ElasticApmTraceId", Agent.Tracer.CurrentTransaction.TraceId));`

The question is still open how to handle `Activity.Current.Id` - it could be added in `GetTracing(LogEvent logEvent)` as fallback. Happy to discuss it here, or in a follow up.

### DIFF
--- a/src/Elastic.Apm.SerilogEnricher/ElasticApmEnricher.cs
+++ b/src/Elastic.Apm.SerilogEnricher/ElasticApmEnricher.cs
@@ -15,9 +15,9 @@ namespace Elastic.Apm.SerilogEnricher
 			if (Agent.Tracer.CurrentTransaction == null) return;
 
 			logEvent.AddPropertyIfAbsent(propertyFactory.CreateProperty(
-				"TransactionId", Agent.Tracer.CurrentTransaction.Id));
+				"ElasticApmTransactionId", Agent.Tracer.CurrentTransaction.Id));
 			logEvent.AddPropertyIfAbsent(propertyFactory.CreateProperty(
-				"TraceId", Agent.Tracer.CurrentTransaction.TraceId));
+				"ElasticApmTraceId", Agent.Tracer.CurrentTransaction.TraceId));
 		}
 	}
 }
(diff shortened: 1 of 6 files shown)


gregkalapos:
I opened https://github.com/elastic/ecs-dotnet/issues/13 to discuss where to put the trace id - as it seems it is currently not where it should be.

codebrain:
Thanks for raising #13. I will open a PR shortly to address this.

codebrain:
I've merged in https://github.com/elastic/ecs-dotnet/pull/16, so after a rebase the properties should now be correct.

gregkalapos:
> I've merged in #16, so after a rebase the properties should now be correct.

Awesome, thanks @codebrain! 

I can verify that #16 made it work. This gif shows jumping from APM to logs and then from a log to APM. 🎉 


My-Movie2

gregkalapos:
One last thing before I merge this:

Are we ok with the new property ids I use in this PR?
- `ElasticApmTransactionId`
- `ElasticApmTraceId `

Given that `TraceId` could be overwritten by Activity.Current. 

@Mpdreamz @codebrain

Mpdreamz:
I am okay with the name, this way we can be sure to apply our our precedence rules 👍

PR #19, opened 2019-12-16
Access to log event during custom mapping

alaw-emoney:
* Ability to enrich the ECS from the log event
* This is a breaking change but not sure if anyone was using this

### DIFF
--- a/src/Elastic.CommonSchema.Serilog/EcsTextFormatterConfiguration.cs
+++ b/src/Elastic.CommonSchema.Serilog/EcsTextFormatterConfiguration.cs
@@ -9,13 +9,14 @@ using System.Web;
 #endif
 using System;
 using System.Runtime.CompilerServices;
+using Serilog.Events;
 
 namespace Elastic.CommonSchema.Serilog
 {
 	public interface IEcsTextFormatterConfiguration
 	{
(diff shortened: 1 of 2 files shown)


alaw-emoney:
Would close issue #20

codebrain:
Hi @alaw-emoney,

Thank you for your contribution, makes complete sense to pass the `LogEvent` to the Func. Would you mind signing our contributor agreement first, so that I am able to merge in this PR?

https://www.elastic.co/contributor-agreement

Thanks

elasticcla:
Hi @alaw-emoney, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/ecs-dotnet/pull/19.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

alaw-emoney:
I have requested someone from eMoney sign the agreement for the whole company

codebrain:
Hi @alaw-emoney - how are you getting on with the CLA signature? I was hoping to release a new package this week and would really like to include this PR change as I believe its very useful. Despite us being in pre-release I would prefer to limit the number of breaking changes if possible.

Issue #20, opened 2019-12-16
Entire ECS schema is available but no way to fill it up


alaw-emoney:
The formatted log contains data within a hard-coded set of fields, which are very useful data points, but leaves a lot of the schema unused, and instead puts log event data into "_metadata". There appears to be no way to move data from the log event to somewhere in the schema. There should be a hook into the formatter such that more properties of the log event can be out of "_metadata" and into the right place in the schema.

Request:
Given the logging user logs properties (e.g. from context), and the user specifies a mapping from the log event to the schema, the formatted text should have the additional schema fields populated.

Mpdreamz:
Now implemented as of #24 we'll do another release shortly. 

Thanks for the PR and issue @alaw-emoney 👍

PR #24, opened 2019-12-19
Change mapping function to map data from the log event

alaw-emoney:
Ability to enrich the ECS from the log event
This is a breaking change but not sure if anyone was using this.
Fixes issue #20


alaw-emoney:
I closed a similar pull request. In the other PR I was logged in as the wrong account when committing the change. Looks like you have an issue in your CI build script?

codebrain:
@alaw-emoney thanks for re-opening and signing the CLA, I will merge this in now.

In terms of the build script, we shall look into it, thanks.

PR #34, opened 2020-01-23
Lock versioning to MAJOR.MINOR.0 instead of MAJOR.0.0

codebrain:
Locking assembly identity to minor versions to account for changes in ECS that would break SemVer for the generated classes.

### DIFF
--- a/build/scripts/Versioning.fs
+++ b/build/scripts/Versioning.fs
@@ -18,7 +18,7 @@ module Versioning =
 
     let parse (v:string) = SemVer.parse(v)
 
-    //Versions in form of e.g 1.0.0 is inferred as datetime so we bake the json shape into the provider like this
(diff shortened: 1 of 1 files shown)


build/scripts/Versioning.fs
```diff
@@ -48,14 +48,15 @@ module Versioning =
...
+
     type BuildVersions =
         | Update of New: AnchoredVersion * Old: AnchoredVersion 
         | NoChange of Current: AnchoredVersion
         
     type ArtifactsVersion = ArtifactsVersion of AnchoredVersion
 
     let AnchoredVersion version = 
-        let av v = parse (sprintf "%s.0.0" (v.Major.ToString()))
+        let av v = parse (sprintf "%s.%s.0" (v.Major.ToString()) (v.Minor.ToString()))
```
russcam: Use `%i` in the format string and remove the `.ToString()` calls

PR #49, opened 2020-02-24
ci(jenkins): add linux build stage

v1v:
## What

Enable the build and test in Linux using a docker container to help with the required dependencies.

## Follow-ups
There will be some other PRs to enable:
- Windows build
- test output visualization in the CI Pipeline.

Depends on https://github.com/elastic/ecs-dotnet/pull/50


.ci/Jenkinsfile
```diff
@@ -47,10 +47,44 @@ pipeline {
         }
       }
     }
+    stage('Parallel'){
```
kuisathaverat: I guess it would be more stages in the future to run in parallel, isn't it?

v1v: Yes, I initially created all the stages there, but I reverted it and created this PR with only the Linux stage for the time being.

PR #50, opened 2020-02-24
ci(jenkins): enable CI Pipeline and JJBB

v1v:
## What

Enable the JJBB and a CI Pipeline to run:

- checkout
- pre-commit validation

There will be some follow-ups to add more stages to the CI Pipeline. See https://github.com/elastic/ecs-dotnet/pull/49

### DIFF
--- a/.ci/jobs/defaults.yml
+++ b/.ci/jobs/defaults.yml
@@ -1,0 +1,59 @@
+---
+
+##### GLOBAL METADATA
+
+- meta:
+    cluster: apm-ci
+
+##### JOB DEFAULTS
+
+- job:
+    view: APM-CI
+    project-type: multibranch
+    logrotate:
+      daysToKeep: 30
+      numToKeep: 100
+    number-to-keep: '5'
+    days-to-keep: '1'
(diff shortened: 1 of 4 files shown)


.ci/jobs/defaults.yml
```diff
@@ -0,0 +1,59 @@
...
+        - change-request:
+            ignore-target-only-changes: false
+        clean:
+          after: true
+          before: true
+        prune: true
+        shallow-clone: true
+        depth: 3
+        do-not-fetch-tags: true
+        submodule:
```
cachedout: Do you think it would be good to add a `reference-repo` here?

v1v: Good point!
The repo is not as big as other git repos and the number of CI builds might not be too much, so I'd say we don't need it for now. and if we need it then we need to add this repo to the list of the reference repos supported in the CI workers.

cachedout: Works for me. 👍

codebrain:
Thanks all.

PR #65, opened 2020-03-19
Added Beta Notice

philkra:
Beta notice to state that the library is subject to potential breaking changes in the future.

Parent ticket: https://github.com/elastic/ecs-logging/issues/24


README.md
```diff
@@ -10,6 +10,12 @@ The Elastic Common Schema defines a common set of fields for ingesting data into
 
 Contributions are welcome, please read our [guidelines](https://github.com/elastic/ecs-dotnet/tree/master/contributing.md).
 
+---
+
+**Please note** that this library is in a **beta** version and backwards-incompatible changes might be introduced in future releases. While we strive to comply to [semver](https://semver.org/), we can not guarantee to avoid breaking changes in minor releases.
```
codebrain:
```suggestion
### Versioning

Version components: `{major}.{minor}.{patch}`

These libraries are not versioned according to [SemVer](https://semver.org/) principles. Backwards compatibility is only guaranteed within *minor* versions, since ECS only makes this guarantee. Patch releases of this library will not seek to introduce breaking changes, but will be used to address bug fixes within that minor version.

Each assembly release indicates the ECS version that it is compatible with (see _releases_), but typically the minor version number of the assembly correlates to the compatible version of ECS; for example; all of `1.4.0`, `1.4.1`, `1.4.2` and `1.4.3` are compatible with ECS version `1.4.0`.

The assemblies are versioned using an assembly identity of `major.minor.*` as opposed to `major.*` as is common when following SemVer.
```

codebrain: The versioning is a little more subtle than the original text, so I have updated with a clearer description of what we do in this repo.

codebrain: https://www.elastic.co/guide/en/ecs/current/ecs-reference.html#_maturity  probably needs updating?

codebrain: Ping @philkra

Issue #61, opened 2020-03-06
Increase elastic apm dependency version in Elastic.Apm.SerilogEnricher


gregkalapos:
In order to fix https://github.com/elastic/ecs-dotnet/issues/58 I opened https://github.com/elastic/apm-agent-dotnet/pull/753 in the .NET APM Agent repo.

That'll fix the issue, but we also have to make sure that `Elastic.Apm.SerilogEnricher` forces users to use the latest agent, otherwise they'll run into the same bug.

So, once the next APM .NET Agent version is out which contains https://github.com/elastic/apm-agent-dotnet/pull/753, we need to update [Elastic.Apm.SerilogEnricher.csproj](https://github.com/elastic/ecs-dotnet/blob/master/src/Elastic.Apm.SerilogEnricher/Elastic.Apm.SerilogEnricher.csproj#L10) and make sure Elastic.Apm.SerilogEnricher depends on the version with the fix.

codebrain:
Makes sense @gregkalapos , if you want to ping in here when it is ready I can handle the increment and publishing.

We have some releases in the pipeline for:

`1.4.3` - https://github.com/elastic/ecs-dotnet/pulls?q=is%3Apr+is%3Aopen+label%3Av1.4.3
`1.5.0` - https://github.com/elastic/ecs-dotnet/pulls?q=is%3Apr+is%3Aopen+label%3Av1.5.0

So will need to ensure this is coordinated properly.

ffMathy:
I guess it is ready now right?

gregkalapos:
Right, Elastic APM 1.4.0 is out which has the fix.
I opened #66.

Issue #58, opened 2020-03-04
[BUG] StackOverflowException when WithElasticApmCorrelationInfo() is used


ffMathy:
**ECS version**: latest as of writing

**Elasticsearch version (if applicable)**: not applicable

**.NET framework / OS**: .NET Core 3.1, Windows and Linux

**Description of the problem, including expected versus actual behavior**:
When using Serilog and the `WithElasticApmCorrelationId` method, I get a `StackOverflowException`. Without that call, I don't get the exception.

Here's a stack trace (which clearly shows there's a lot of recursion).

> 	System.Private.CoreLib.dll!System.Globalization.CompareInfo.FindString(uint dwFindNLSStringFlags, string lpStringSource, int startSource, int cchSource, string lpStringValue, int startValue, int cchValue, int* pcchFound)	Unknown
>  	System.Private.CoreLib.dll!System.Globalization.CompareInfo.StartsWith(string source, string prefix, System.Globalization.CompareOptions options)	Unknown
>  	System.Private.CoreLib.dll!System.Globalization.CompareInfo.IsPrefix(string source, string prefix, System.Globalization.CompareOptions options)	Unknown
>  	System.Private.CoreLib.dll!string.StartsWith(string value, System.StringComparison comparisonType)	Unknown
>  	System.Private.CoreLib.dll!string.StartsWith(string value)	Unknown
>  	Serilog.Extensions.Logging.dll!Serilog.Extensions.Logging.SerilogLogger.Log<Elastic.Apm.Logging.LogValuesFormatter.LogValues>(Microsoft.Extensions.Logging.LogLevel logLevel, Microsoft.Extensions.Logging.EventId eventId, Elastic.Apm.Logging.LogValuesFormatter.LogValues state, System.Exception exception, System.Func<Elastic.Apm.Logging.LogValuesFormatter.LogValues, System.Exception, string> formatter)	Unknown
>  	Elastic.Apm.AspNetCore.dll!Elastic.Apm.AspNetCore.AspNetCoreLogger.Log<Elastic.Apm.Logging.LogValuesFormatter.LogValues>(Elastic.Apm.Logging.LogLevel level, Elastic.Apm.Logging.LogValuesFormatter.LogValues state, System.Exception e, System.Func<Elastic.Apm.Logging.LogValuesFormatter.LogValues, System.Exception, string> formatter)	Unknown
[...]

PR #66, opened 2020-03-23
Increase Elastic.Apm dependency to 1.4.0

gregkalapos:
Solves #61.

With this by default we'll have a version out which contains a fix to https://github.com/elastic/ecs-dotnet/issues/58.


ffMathy:
I don't see what could go wrong releasing this (famous last words).

No but seriously.

codebrain:
Looks good, will get this merged in and a new set of packages released.

codebrain:
Released on nuget in version `1.4.4`

PR #67, opened 2020-03-30
EcsLayout - Use LogEventInfo.Logger instead of hardcoded DefaultLogger

snakefoot:
Resolves #56

PR #74, opened 2020-05-07
NLog Ecs Layout event Enrichment

ghayes7:
This PR allows for both programmatical and derivational enrichment of NLog ecs events before they are serialized.

Example:
`var tt = new EcsLayout()
{
	EnrichAction = (ecsEvent,logEvent) =>
	{
		if (ecsEvent.Event != null)
			ecsEvent.Event.Dataset = "someValue";
	}
}`

Or deriving a class from NLog.EcsLayout and overriding the EnrichEvent function.

Both ways are not needed but at least one is. Otherwise a class can't event be derived from [EcsLayout ](https://github.com/elastic/ecs-dotnet/blob/bbc0ff2cd25219e58b2f27956495e2bb8f314277/src/Elastic.CommonSchema.NLog/EcsLayout.cs#L20) and override the [RenderFormattedMessage ](https://github.com/elastic/ecs-dotnet/blob/bbc0ff2cd25219e58b2f27956495e2bb8f314277/src/Elastic.CommonSchema.NLog/EcsLayout.cs#L98) function because all of the called Get* functions are marked as private (so they too would have to be cloned). The only option, would be to derive, override the RenderFormattedMessage function, call the base implementation with a clean stringbuilder, deserialize the return, make changes, and reserialize... Which would just be silly...

### DIFF
--- a/src/Elastic.CommonSchema.NLog/EcsLayout.cs
+++ b/src/Elastic.CommonSchema.NLog/EcsLayout.cs
@@ -86,10 +86,11 @@ namespace Elastic.CommonSchema.NLog
 		public Layout ServerAddress { get; set; }
 		public Layout ServerIp { get; set; }
 		public Layout ServerUser { get; set; }
+
 		/// <summary>
-        /// Optional action to enrich the constructed <see cref="Base">EcsEvent</see> before it is serialized
-        /// </summary>
-        /// <remarks>This is called last in the chain of enrichment functions</remarks>
+		/// Optional action to enrich the constructed <see cref="Base">EcsEvent</see> before it is serialized
+		/// </summary>
+		/// <remarks>This is called last in the chain of enrichment functions</remarks>
 		public Action<Base,LogEventInfo> EnrichAction { get; set; }
 
 		[ArrayParameter(typeof(TargetPropertyWithContext), "tag")]
@@ -124,12 +125,12 @@ namespace Elastic.CommonSchema.NLog
 		}
 
 		/// <summary>
-        /// Override to supplement the ECS event parsing
-        /// </summary>
-        /// <param name="logEventInfo">The original log event</param>
-        /// <param name="ecsEvent">The EcsEvent to modify</param>
-        /// <returns>Enriched ECS Event</returns>
-        /// <remarks>Destructive for performance</remarks>
+		/// Override to supplement the ECS event parsing
+		/// </summary>
+		/// <param name="logEventInfo">The original log event</param>
+		/// <param name="ecsEvent">The EcsEvent to modify</param>
+		/// <returns>Enriched ECS Event</returns>
+		/// <remarks>Destructive for performance</remarks>
 		protected virtual void EnrichEvent(LogEventInfo logEventInfo,ref Base ecsEvent)
 		{
 		}
--- a/src/Elastic.CommonSchema.NLog/EcsLayout.cs
+++ b/src/Elastic.CommonSchema.NLog/EcsLayout.cs
@@ -91,6 +91,11 @@ namespace Elastic.CommonSchema.NLog
 		public Layout ServerAddress { get; set; }
 		public Layout ServerIp { get; set; }
 		public Layout ServerUser { get; set; }
+		/// <summary>
+        /// Optional action to enrich the constructed <see cref="Base">EcsEvent</see> before it is serialized
+        /// </summary>
+        /// <remarks>This is called last in the chain of enrichment functions</remarks>
+		public Action<Base,LogEventInfo> EnrichAction { get; set; }
 
 		[ArrayParameter(typeof(TargetPropertyWithContext), "tag")]
 		public IList<TargetPropertyWithContext> Tags { get; } = new List<TargetPropertyWithContext>();
@@ -115,9 +120,23 @@ namespace Elastic.CommonSchema.NLog
 				Server = GetServer(logEventInfo),
 				Host = GetHost(logEventInfo)
 			};
-
+			//Give any deriving classes a chance to enrich the event
+			EnrichEvent(logEventInfo,ref ecsEvent);
+			//Allow programmatical actions to enrich before serializing
+			EnrichAction?.Invoke(ecsEvent, logEventInfo);
 			var output = ecsEvent.Serialize();
 			target.Append(output);
+		}
+
+		/// <summary>
+        /// Override to supplement the ECS event parsing
+        /// </summary>
+        /// <param name="logEventInfo">The original log event</param>
+        /// <param name="ecsEvent">The EcsEvent to modify</param>
+        /// <returns>Enriched ECS Event</returns>
+        /// <remarks>Destructive for performance</remarks>
+		protected virtual void EnrichEvent(LogEventInfo logEventInfo,ref Base ecsEvent)
+		{
 		}
 
 		protected override string GetFormattedMessage(LogEventInfo logEvent)


ghayes7:
Signed the contributor agreement. Do I need to re-submit the PR?

snakefoot:
@ghayes7 NLog really likes the Layout logic, as it ensures correct context capture whether used in sync or async mode. Also the Layout logic is easy to configure from config-files.

Can you give an example where you would use this enricher? And where the Layout logic is failing?

ghayes7:
> @ghayes7 NLog really likes the Layout logic, as it ensures correct context capture whether used in sync or async mode. Also the Layout logic is easy to configure from config-files.
> 
> Can you give an example where you would use this enricher? And where the Layout logic is failing?

@snakefoot Agreed in regards to NLog's affinity for layouts. The problem with the Elastic.CommonSchema.NLog.EcsLayout (hereinafter referred to as EcsLayout) is that it only captures an extremely small subset of fields in the ECS; which is fine for the common use audience. However, for those that need more, there are no practical paths to extend or supplement the small subset of supported ECS fields in the EcsLayout. The Elasitc.CommonSchema.Serilog at least has the [MapCustom ](https://github.com/elastic/ecs-dotnet/blob/bbc0ff2cd25219e58b2f27956495e2bb8f314277/src/Elastic.CommonSchema.Serilog/EcsTextFormatterConfiguration.cs#L44) prop to do this.

So of the two possible approaches this PR enables (derivative and programmatical):
With derived classes you could extend and specify needed field's, (e.g. event.dataset and event.module), give them an NLog layout and override EcsLayout.Enrich (off the top of my head, please forgive syntax issues):
```C#
public class EcsDerivedLayout : EcsLayout{
    public Layout EventDataset {get; set;}
    public Layout EventModule {get; set;}
    protected virtual void EnrichEvent(LogEventInfo logEventInfo,ref Base ecsEvent)
    {
        ecsEvent.Event.Dataset= EventDataset?.Render(logEventInfo);
        ecsEvent.Event.Module= EventModule?.Render(logEventInfo);
    }
}
```
(Of course with some null checking and what have you)
Whereas the programmatical way (i.e. specifying an action for EcsLayout.EnrichAction => ...) can be used for program static & global overrides that'd not be worth the trouble going through the Layout .route (various fields in: Agent, OS, Server, Service, etc). Again, not terribly applicable, but could also be used similar to the above only using external, static fields in place of the class props:
```C#
internal static class EcsFieldConstants{}
    //Layout Renderer's, registred elsewhere...
    public static Layout EventDataset {get; set;} = "${ecsEventDataset}";
    public static Layout EventModule {get; set;} = "${ecsEventModule}";
}
public static void BuildLayout(){    
    var config = new Config.LoggingConfiguration();
    var layout = new EcsLayout() { 
        EnrichAction = (ecsEvent,logEvent) => 
        { 
            if (ecsEvent.Event != null){
                ecsEvent.Event.Dataset = EcsFieldConstants.EventDataset?.Render(logEvent); 
                ecsEvent.Event.Module = EcsFieldConstants.EventModule?.Render(logEvent); 
            }
                
        }; 
    }

    var memoryTarget = new EventInfoMemoryTarget { Layout = layout }; // Use the layout.
    config.AddRule(LogLevel.Debug, LogLevel.Fatal, memoryTarget);
    var factory = new LogFactory(config);
    var logger = factory.GetCurrentClassLogger();
    
}
```
With the correlating LayoutRenderers registered elsewhere. Again, less eloquent than the derived class example, but a feasible path.

Neither are terribly NLog config friendly... You could theoretically add a dictionary of additional fields and source it from the config, then resolve them to the ECS.Base class fields using reflection (e.g. "event.module" mapping to base.event.module with value of either a layout or a const)... However it'd not be very efficient.

snakefoot:
Like your combination of `EnrichEvent`-method with public  `Layout`-properties, then it will work okay:

```c#
public class EcsDerivedLayout : EcsLayout{
    public Layout EventDataset {get; set;}
    public Layout EventModule {get; set;}
    protected virtual void EnrichEvent(LogEventInfo logEventInfo,ref Base ecsEvent)
    {
        ecsEvent.Event.Dataset= EventDataset?.Render(logEventInfo);
        ecsEvent.Event.Module= EventModule?.Render(logEventInfo);
    }
}
```

Be vary that  #67 conflicts with this PR, as it removes override of `RenderFormattedMessage`.

I was just thinking that common properties used by many should be included by default in EcsLayout. Instead of everyone doing their own story.

ghayes7:
> Be vary that #67 conflicts with this PR, as it removes override of `RenderFormattedMessage`.

Looks like you fixed it with: db459e9? (thanks)

Might make more sense to move the process of building the ecs event from the log event,to a seperate, overridable function... i.e.
```C#
protected virtual Base BuildEcsBase(LogEventInfo logEventInfo)
{
    var ecsEvent = new Base
    {
        Timestamp = logEventInfo.TimeStamp,
        Message = logEventInfo.FormattedMessage,
        Ecs = new Ecs { Version = Base.Version },
        Log = GetLog(logEventInfo),
        Event = GetEvent(logEventInfo),
        //...
        Host = GetHost(logEventInfo)
    };

    return ecsEvent;
}
protected override void RenderFormattedMessage(LogEventInfo logEventInfo, StringBuilder target)
{
    var ecsEvent = BuildEcsBase()

    var output = ecsEvent.Serialize();
    target.Append(output);
}

```
Then in a derived class I Just:
```C#
public class EcsDerivedLayout: EcsLayout{
    protected override Base BuildEcsBase(LogEventInfo logEventInfo)
    {
        var ecsEvent = BuildEcsBase(logEventInfo);
        //Do stuff to ecsEvent
        return ecsEvent;
    }
}
```
And inject my own fields, exactly like in my previous comment with the ecs fields I need as class props of type layout.

> I was just thinking that common properties used by many should be included by default in EcsLayout. Instead of everyone doing their own story.

100% agree, but need a solution in the interim. Both the NLog and Serilog ECS libraries have extremely small coverage of the ECS. Though there's an argument to be made that the data source (in NLog's case, the LayoutRenderers) of many of the less common, but still pertinent fields that are not currently covered, probably don't have an existing renderer. Though at least, partial support can be added (i.e. adding the correlating fields as props of type Layout, but kept null, to leave implementation up to others).

Regardless, If someone expects others to use their proposed standard, then they need to enable the full use of that standard and not just the limited portion they believe to be pertinent. Or at the very least, allow extensibility to add coverage as needed (i.e. not marking functions as private so derived classes cannot build off of them).

Though, it does logically hold that if the subset of fields supported, do cover all logging uses, then any additional fields would be a custom implementation that wouldn't hold a common use case to warrant commiting to the public repo. For example, if I were to want to use logging as a means for Audit logging for my application. Fields that I'd extract for the various correlative ECS fields may be specific to my application... For example User, Organization, network, etc. Say we have some form of service mesh or related proxy. How I'd extract that information at a given point (say my apps a web api, so somewhere in the httpcontext) may be very different for other applications.

snakefoot:
> Looks like you fixed it 

Yes I discovered it was possible to write directly into StringBuilder without constant memory-allocation. So will keep `RenderFormattedMessage`.

> Both the NLog and Serilog ECS libraries have extremely small coverage of the ECS

Yes that also confuses me. But maybe they are just kicking the tires to see what happens, and then later complete the prototype.

>  Fields that I'd extract for the various correlative ECS fields may be specific to my application... For example User, Organization, network, etc. Say we have some form of service mesh or related proxy. How I'd extract that information at a given point (say my apps a web api, so somewhere in the httpcontext) may be very different for other applications.

This is where NLog Layout or Serilog Enrichers should be used. They act as mapping logic/layer/helpers between the application-context to the final output.

Mpdreamz:
I am less experienced with NLog so this PR initially looked good to me as the analogous of MapCustom in the Serilog provider. 

> Both the NLog and Serilog ECS libraries have extremely small coverage of the ECS

This was indeed on purpose, for Serilog/NLog we tried to support the most common enrichers out of the box but did not go out of our way to support all enrichers out there.

> I was just thinking that common properties used by many should be included by default in EcsLayout. Instead of everyone doing their own story.

This would be ideal, happy to hear more where we can fill in the blanks out of the box better,

> > Fields that I'd extract for the various correlative ECS fields may be specific to my application... For example User, Organization, network, etc. Say we have some form of service mesh or related proxy. How I'd extract that information at a given point (say my apps a web api, so somewhere in the httpcontext) may be very different for other applications.

> This is where NLog Layout or Serilog Enrichers should be used. They act as mapping logic/layer/helpers between the application-context to the final output.

This is only partially true, for generic enrichers/layout parameters values end up in `metadata`. This a catch all bucket that ideally should be avoided if the values map better to `ECS` fields. 

This PR would allow users to map their own correlation to proper ECS fields.

src/Elastic.CommonSchema.NLog/EcsLayout.cs
```diff
@@ -86,6 +86,11 @@ public EcsLayout()
 		public Layout ServerAddress { get; set; }
 		public Layout ServerIp { get; set; }
 		public Layout ServerUser { get; set; }
+		/// <summary>
+        /// Optional action to enrich the constructed <see cref="Base">EcsEvent</see> before it is serialized
```
Mpdreamz, line 90: Mind changes the spaces to tabs on this PR?

ghayes7: Sure thing! Done with: https://github.com/ghayes7/ecs-dotnet/commit/da34b9dfa4fa1c55254066af5de61ddff9e7e50a

Mpdreamz:
Jenkins run the tests please

Mpdreamz:
Thanks a ton for the PR @ghayes7 👍 

I won't look at preparing a new release till next week earliest though but looking forward to highlighting this in the release notes 😄

PR #75, opened 2020-05-08
NLog EcsLayout - Write directly to StringBuilder instead of string allocation

snakefoot:
Also recycles the same Utf8JsonWriter when possible (It is intended to be reused see also [Utf8JsonWriter.Reset](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonwriter.reset))

PR #82, opened 2020-06-02
Expose Serialize(StringBuilder) overload

Mpdreamz:
Continuation of #75 

* NLog EcsLayout - Write directly to StringBuilder instead of string allocation

This continuation:

- Move internal ReusableUtf8JsonWriter down to Elastic.CommonSchema
- Expose an overload on `Base` that serializes on a `StringBuilder` directly.
- Add benchmarks for the `StringBuilder` overload
- `ReusableUtf8JsonWriter` is now a `ThreadStatic` instance on `Base`

Reusing the text writer in the `StringBuilder` overload:

|  Method |       Mean |      Error |      StdDev | Completed Work Items | Lock Contentions |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------- |-----------:|-----------:|------------:|---------------------:|-----------------:|-------:|------:|------:|----------:|
|   Empty |   1.064 μs |  0.0047 μs |   0.0042 μs |               0.0000 |                - | 0.1068 |     - |     - |     680 B |
| Minimal |   1.740 μs |  0.0067 μs |   0.0059 μs |               0.0000 |                - | 0.1431 |     - |     - |     903 B |
| Complex |   4.732 μs |  0.0219 μs |   0.0183 μs |               0.0000 |                - | 0.3357 |     - |     - |    2120 B |
|    Full | 484.817 μs | 39.2360 μs | 112.5753 μs |                    - |                - |      - |     - |     - |   58208 B |

vs newing one up every time in the `StringBuilder` overload.

|  Method |       Mean |      Error |     StdDev | Completed Work Items | Lock Contentions |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------- |-----------:|-----------:|-----------:|---------------------:|-----------------:|-------:|------:|------:|----------:|
|   Empty |   1.112 μs |  0.0047 μs |  0.0044 μs |               0.0000 |                - | 0.2098 |     - |     - |   1.29 KB |
| Minimal |   1.808 μs |  0.0032 μs |  0.0028 μs |               0.0000 |                - | 0.2193 |     - |     - |   1.34 KB |
| Complex |   4.803 μs |  0.0267 μs |  0.0236 μs |               0.0000 |                - | 0.4272 |     - |     - |   2.66 KB |
|    Full | 492.265 μs | 32.6823 μs | 93.7718 μs |                    - |                - |      - |     - |     - |  45.94 KB |

For reference this is the unmodified `ecs.Serialize()` path which returns a string

|  Method |       Mean |      Error |     StdDev |     Median | Completed Work Items | Lock Contentions |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------- |-----------:|-----------:|-----------:|-----------:|---------------------:|-----------------:|-------:|------:|------:|----------:|
|   Empty |   1.092 μs |  0.0096 μs |  0.0090 μs |   1.090 μs |               0.0000 |                - | 0.1125 |     - |     - |     712 B |
| Minimal |   1.766 μs |  0.0049 μs |  0.0044 μs |   1.765 μs |               0.0000 |                - | 0.1469 |     - |     - |     928 B |
| Complex |   4.805 μs |  0.0342 μs |  0.0320 μs |   4.805 μs |               0.0000 |                - | 0.3357 |     - |     - |    2152 B |
|    Full | 480.523 μs | 34.5599 μs | 99.7134 μs | 445.317 μs |                    - |                - |      - |     - |     - |   41872 B |

Which seems to have on par performance with the stringbuilder overload that reuses.

### DIFF
--- a/src/Elastic.CommonSchema/Base.Serialization.cs
+++ b/src/Elastic.CommonSchema/Base.Serialization.cs
@@ -106,6 +106,8 @@ namespace Elastic.CommonSchema
 			JsonSerializer.Serialize(writer, this, JsonConfiguration.SerializerOptions);
 		}
 
+		public void Serialize(Utf8JsonWriter writer) => JsonSerializer.Serialize(writer, this, JsonConfiguration.SerializerOptions);
+
 		public Task SerializeAsync(Stream stream, CancellationToken ctx = default) =>
 			JsonSerializer.SerializeAsync(stream, this, GetType(), SerializerOptions, ctx);
 	}
--- a/src/Elastic.CommonSchema/Base.Serialization.cs
+++ b/src/Elastic.CommonSchema/Base.Serialization.cs
@@ -98,8 +98,9 @@ namespace Elastic.CommonSchema
 		public byte[] SerializeToUtf8Bytes() => JsonSerializer.SerializeToUtf8Bytes(this, GetType(), SerializerOptions);
 
 		[ThreadStatic]
-		private static readonly ReusableUtf8JsonWriter ReusableJsonWriter = new ReusableUtf8JsonWriter();
-		
+		private static ReusableUtf8JsonWriter _reusableJsonWriter;
+		private static ReusableUtf8JsonWriter ReusableJsonWriter => _reusableJsonWriter ??= new ReusableUtf8JsonWriter();
+
 		public StringBuilder Serialize(StringBuilder stringBuilder)
 		{
 			using var reusableWriter = ReusableJsonWriter.AllocateJsonWriter(stringBuilder);
--- a/src/Elastic.CommonSchema/Base.Serialization.cs
+++ b/src/Elastic.CommonSchema/Base.Serialization.cs
@@ -3,7 +3,6 @@
 // See the LICENSE file in the project root for more information
 
 using System;
-using System.Buffers;
 using System.IO;
 using System.Text;
 using System.Text.Encodings.Web;
@@ -12,45 +11,6 @@ using System.Threading;
 using System.Threading.Tasks;
 using Elastic.CommonSchema.Serialization;
 using static Elastic.CommonSchema.Serialization.JsonConfiguration;
-
-namespace Elastic.CommonSchema.Serialization
-{
-	/// <summary>
-	/// This static class allows you to deserialize subclasses of <see cref="Base"/>
-	/// If you are dealing with <see cref="Base"/> directly you do not need to use this class,
-	/// use <see cref="Base.Deserialize(string)"/> and the overloads instead.
-	/// </summary>
-	/// <remarks>
-	/// This class should only be used for advanced use cases, for simpler use cases you can utilise the <see cref="Base.Metadata"/> property.
-	/// </remarks>
-	/// <typeparam name="TBase">Type of the <see cref="Base"/> subclass</typeparam>
-	public static class EcsSerializerFactory<TBase> where TBase : Base, new()
-	{
-		public static ValueTask<TBase> DeserializeAsync(Stream stream, CancellationToken ctx = default) =>
-			JsonSerializer.DeserializeAsync<TBase>(stream, SerializerOptions, ctx);
-
-		public static TBase Deserialize(string json) => JsonSerializer.Deserialize<TBase>(json, SerializerOptions);
-
-		public static TBase Deserialize(ReadOnlySpan<byte> json) => JsonSerializer.Deserialize<TBase>(json, SerializerOptions);
-
-		public static TBase Deserialize(Stream stream)
-		{
-			using var ms = new MemoryStream();
-			var buffer = ArrayPool<byte>.Shared.Rent(1024);
-			var total = 0;
-			int read;
-			while ((read = stream.Read(buffer, 0, buffer.Length)) > 0)
-			{
-				ms.Write(buffer, 0, read);
-				total += read;
-			}
-			var span = ms.TryGetBuffer(out var segment)
-				? new ReadOnlyMemory<byte>(segment.Array, segment.Offset, total).Span
-				: new ReadOnlyMemory<byte>(ms.ToArray()).Span;
-			return Deserialize(span);
-		}
-	}
-}
 
 namespace Elastic.CommonSchema
 {
@@ -97,7 +57,6 @@ namespace Elastic.CommonSchema
 
 		public byte[] SerializeToUtf8Bytes() => JsonSerializer.SerializeToUtf8Bytes(this, GetType(), SerializerOptions);
 
-		[ThreadStatic]
 		private static ReusableUtf8JsonWriter _reusableJsonWriter;
 		private static ReusableUtf8JsonWriter ReusableJsonWriter => _reusableJsonWriter ??= new ReusableUtf8JsonWriter();
 
--- a/src/Elastic.CommonSchema/Base.Serialization.cs
+++ b/src/Elastic.CommonSchema/Base.Serialization.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Buffers;
 using System.IO;
+using System.Text;
 using System.Text.Encodings.Web;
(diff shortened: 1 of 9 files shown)


src/Elastic.CommonSchema/Base.Serialization.cs
```diff
@@ -96,16 +97,29 @@ protected internal virtual bool TryRead(string propertyName, out Type type)
 
 		public byte[] SerializeToUtf8Bytes() => JsonSerializer.SerializeToUtf8Bytes(this, GetType(), SerializerOptions);
 
+		[ThreadStatic]
+		private static ReusableUtf8JsonWriter _reusableJsonWriter;
```
snakefoot, line 101: Must admit I really dislike `[ThreadStatic]`-objects. Why not let ReusableUtf8JsonWriter live in the NLog-project, until others needs it?

src/Elastic.CommonSchema/Base.Serialization.cs
```diff
@@ -96,16 +97,29 @@ protected internal virtual bool TryRead(string propertyName, out Type type)
...
+		private static ReusableUtf8JsonWriter ReusableJsonWriter => _reusableJsonWriter ??= new ReusableUtf8JsonWriter();
+
+		public StringBuilder Serialize(StringBuilder stringBuilder)
+		{
+			using var reusableWriter = ReusableJsonWriter.AllocateJsonWriter(stringBuilder);
+			reusableWriter.Serialize(this);
+			return stringBuilder;
```
snakefoot, line 108: Any reason for having the StringBuilder as return parameter?

tests/Elastic.CommonSchema.Benchmarks/SerializingStringBuilderBase.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace Elastic.CommonSchema.Benchmarks
+{
+	[UnicodeConsoleLogger, MemoryDiagnoser, ThreadingDiagnoser]
+	public class SerializingStringBuilderBase
+	{
+		[Benchmark]
+		public StringBuilder Empty()
+		{
+			var ecs = new Base();
+			return ecs.Serialize(new StringBuilder());
```
snakefoot, line 15: Curious about these benchmarks, since the whole idea is to test the performance of reusing the same StringBuilder, instead of allocating new strings. Would think the StringBuilder should be a member-variable initialized outside the benchmark-code.

src/Elastic.CommonSchema/Base.Serialization.cs
Mpdreamz: Makes the method more composable e.g with linq's `.Aggregate()`

tests/Elastic.CommonSchema.Benchmarks/SerializingStringBuilderBase.cs
Mpdreamz: In the NLog integration we get a `StringBuilder` from `RenderFormattedMessage` which as far as I can tell is not guaranteed to be passed a cached/reused `StringBuilder` either.

src/Elastic.CommonSchema/Base.Serialization.cs
snakefoot: Not besides the clutter of having the object allocated for every threadpool thread, and the inability to release the object unless the thread is recycled. Anyway not my library to maintain or use.

tests/Elastic.CommonSchema.Benchmarks/SerializingStringBuilderBase.cs
snakefoot: NLog Targets makes use of an object-pool of StringBuilders. So yes NLog will cache/reuse the same StringBuilder again and again. Which is why it is recommended to write "directly" to StringBuilder instead of allocating strings.

russcam: I think there should be some additional benchmarks to also compare with a reused `StringBuilder`, as that will measure reuse in conjunction with reusing json writer on the same thread.

Mpdreamz: Happy to amend, I went of the worst case scenario:

https://github.com/NLog/NLog/blob/29879ece25a7d2e47a148fc3736ec310aee29465/src/NLog/Layouts/Layout.cs#L278

Where `RenderAllocateBuilder` is called within Nlog without a `reusableBuilder`. We don't have the same infrastructure NLog has with https://github.com/NLog/NLog/blob/86fba05e272a07d508b32475957b9310f607061d/src/NLog/Internal/ObjectPools/StringBuilderPool.cs

The benchmark is not indicative of how fast NLog can go, but rather tests the worst case scenario where someone passes a new StringBuilder to our `Serialize()` method which to me still yields acceptable performance.

The reuse of json writer is not depending on passing the same StringBuilder either.

src/Elastic.CommonSchema/Base.Serialization.cs
Mpdreamz: @russcam what would be the benefit? The way its currently implemented runs the initialization on each thread too. `IDisposable` brings no benefit in static context. 

If having an object allocated for each threadpool thread is an issue we can remove `ThreadStatic` alltogether and run with a single instance for simplicity sake knowing cache misses inside the `ReusableUtf8JsonWriter` are more likely.

This to me is a prime example of the few cases I'd want ThreadStatic as the lifetime is explictly bound to the thread and we always want to have 1 instance per thread. Since this is part of logging which is likely happening throughout the whole lifetime of a thread I'm also somewhat less concerned about releasing.

Since this is contentious wIil remove the `ThreadStatic` for now and we can revisit in the future.

Mpdreamz:
> Thoughts on decoupling serialization from Elastic.CommonSchema , in a separate project?

I would like to collect more evidence this is needed over time. There are several ways the library could evolve, e.g we could generate readers/writers for the types, look in to shading the dependency and make `Elastic.CommonSchema` dependency free. 

Other patterns with `System.Text.Json` revolve around reusable bufferwriters which wrap around `JsonWriters`.

Mpdreamz:
Thanks for the reviews @snakefoot @russcam I'm opting not to include benchmarks for reusable StringBuilders since that includes including a stringbuilder pool implementation not sure what proving that is faster does for us. Happy to take PR's though.

src/Elastic.CommonSchema/Base.Serialization.cs
russcam: > what would be the benefit? The way its currently implemented runs the initialization on each thread too. IDisposable brings no benefit in static context.

`ThreadLocal<ReusableUtf8JsonWriter>` brings declaration and initialization together, so maybe a little easier to reason about? Technically they would both work the same here as far as I can see

russcam:
> I would like to collect more evidence this is needed over time. There are several ways the library could evolve, e.g we could generate readers/writers for the types, look in to shading the dependency and make Elastic.CommonSchema dependency free.

At the moment, serialization behaviour is coupled to the common schema data models. In having this coupling, the library imposes a specific implementation, whether this is using System.Text.Json, ThreadStatic cached JSON writers, etc. that a user of the common schema data models is not able to change or implement differently to suit their needs, because the behaviour is encapsulated in the model.

Undoubtedly, a user will want to serialize their common schema model to JSON at some point (as this is one of the primary goals in providing them 😃), and with serialization decoupled from data, both the library and users would be provided more flexibility in how this serialization behaves, whilst still providing a default. It would also be a move towards making Elastic.CommonSchema dependency free.

snakefoot:
Could also have the factory to return a functor-delegate that captures the reusable-utf8-writer.

The functor takes StringBuilder and Base as Input, and internally makes use of of its own instance of ReusableUtf8JsonWriter

PR #83, opened 2020-06-02
fix #79 only inject layout renderer keys for APM if Elastic.Apm.NLog …

### LINKED ISSUE #79: [BUG] Using NLog a required dependency isn't provded by NuGet
**ECS schema version (e.g. 1.4.0)**:
1.5.0

**ECS .NET assembly version (e.g. 1.4.2)**:
1.5.0

**Elasticsearch version (if applicable)**:
7.2+

**.NET framework / OS**:
windows dotnetcore 3.1

When NLog starts up using a default configured EcsLayout, an exception is thrown that `ElasticApmTraceId` can not be found. This is because `ElasticApmTraceId` and `ElasticApmTransactionId` are defined in Package `Elastic.Apm.NLog`, however `Elastic.Apm.NLog` is not a required dependency of `Elastic.CommonSchema.NLog` through NuGet

**Steps to reproduce**:
1. Add `NLog.Targets.ElasticSearch` (alpha version, see above) and `Elastic.CommonSchema.NLog` 1.5.0 to your project.
2. Configure the project using the minimal options:
```xml
<extensions>
    <add assembly="NLog.Targets.ElasticSearch"/>
    <add assembly="Elastic.CommonSchema.NLog"/>
</extensions>
<targets>
   <target xsi:type="ElasticSearch" enableJsonLayout="true">
       <layout xsi:type="EcsLayout" />
   </target>
</targets>
```
3. Try to write a log message that is routed to the elastic target

exception is thrown:
```
System.TypeInitializationException
  HResult=0x80131534
  Message=The type initializer for 'Program' threw an exception.
  Source=Program
  StackTrace:
   at Program.<Main>d__7.MoveNext() in ....\Program.cs:line 34

  This exception was originally thrown at this call stack:
    [External Code]

Inner Exception 1:
NLogConfigurationException: Exception when parsing ....\bin\Debug\netcoreapp3.1\NLog.config. 

Inner Exception 2:
TargetInvocationException: Exception has been thrown by the target of an invocation.

Inner Exception 3:
ArgumentException: LayoutRenderer cannot be found: 'ElasticApmTraceId'
```

**Expected Behavior**:
Either the NuGet package should list `Elastic.Apm.NLog` as a dependency, or the Layout should be configured that it only uses items out of the `Elastic.Apm.NLog` package when it's available (possibly through reflection or a configuration option)

Mpdreamz:
…is referenced

### DIFF
--- a/src/Elastic.CommonSchema.NLog/EcsLayout.cs
+++ b/src/Elastic.CommonSchema.NLog/EcsLayout.cs
@@ -6,6 +6,8 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
+using System.Reflection;
 using System.Text;
 using NLog;
 using NLog.Config;
@@ -20,6 +22,9 @@ namespace Elastic.CommonSchema.NLog
 	public class EcsLayout : Layout
 	{
 		public const string Name = nameof(EcsLayout);
+
+		private static readonly bool NLogApmLoaded =
+			Type.GetType("Elastic.Apm.NLog.ApmTraceIdLayoutRenderer, Elastic.Apm.NLog") != null
 
 		private readonly Layout _disableThreadAgnostic = "${threadid:cached=true}";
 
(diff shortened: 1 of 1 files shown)


src/Elastic.CommonSchema.NLog/EcsLayout.cs
```diff
@@ -21,6 +23,9 @@ public class EcsLayout : Layout
 	{
 		public const string Name = nameof(EcsLayout);
 
+		private static readonly bool NLogApmLoaded =
```
snakefoot, line 26: Think it should be a static-method instead of being static-member-variable that is initialized on class-type-initiailization. It is recommended to reduce the amount of logic being execute on class-type-initialization, and instead perform the logic in class-object-creation.

Mpdreamz: ++ will amend 👍

PR #84, opened 2020-06-02
fix #68 labels should be dict of string to string

### LINKED ISSUE #68: [BUG] Base labels values should probably be string (keyword), not object
**ECS schema version (e.g. 1.4.0)**:
1.5.0

**ECS .NET assembly version (e.g. 1.4.2)**:
1.5.0

**Elasticsearch version (if applicable)**:

**.NET framework / OS**:
Core 3.1 / Linux (or any)

The Labels property is defined as IDictionary<string, object>, but should probably be IDictionary<string, string>

In the specification / YAML the type of labels is 'object', but I think that refers to the labels item itself, i.e. the dictionary is the object, not the contents.

The text of the specification defines the values as type 'keyword' (i.e string), and specifically states they should not be nested objects. The yaml also has a different property, object_type: keyword, that I think defines the type of the contents.

Allowing objects is obviously cooler, because there is semantic value in distinguishing between "5" and the number 5, etc, but I think the spec is that the values are supposed to be strings.

**Steps to reproduce**:
 1. Add a labels key/value with a scalar, e.g. int, or nested objects:
     logEvent.Labels.Add("Foo", 5);
     logEvent.Labels.Add("bar", new User() { Name = "waz" });

 2. See the JSON in Elastic / Kibana
    "labels": {
      "foo": 6,
      "bar": {
        "name": "waz"
       }
      },

Mpdreamz:
Fixes #68

### DIFF
--- a/src/Elastic.CommonSchema.NLog/EcsLayout.cs
+++ b/src/Elastic.CommonSchema.NLog/EcsLayout.cs
@@ -301,7 +301,7 @@ namespace Elastic.CommonSchema.NLog
 
 		private string[] GetTags(LogEventInfo e)
 		{
-			if (Tags?.Count == 0)
+			if (Tags is null || Tags.Count == 0)
 				return null;
 
 			if (Tags.Count == 1)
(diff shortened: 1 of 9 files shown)


Mpdreamz:
@gregkalapos @russcam @sgryphon this one has to go in `1.6.0` since we a have minor non breaking clause in https://github.com/elastic/ecs-dotnet#versioning

russcam:
Wondering if labels should be `Dictionary<string, string>` or whether it should be a Dictionary constrained to scalar values. For example, the APM agent labels can be any of `string`, `number` or `boolean`, and are modelled internally with a `Label` type:

https://github.com/elastic/apm-agent-dotnet/blob/2c433d31c45444e026081721e9c4190dadc383b8/src/Elastic.Apm/Api/Label.cs

As far as I understand, APM agent labels map to the ECS labels field. Should Elastic Common Schema .NET have a similar type to constrain label values?

sgryphon:
@russcam this fix needs to be in 1.6 (if it is not there already... although I guess you just filter on tag = 1.6)

PR #80, opened 2020-05-28
Update build

Mpdreamz:
This updates the build of this repos to be easier in the long run. 

Build scripts are simplified. 

Pushes to feedz.org for canary nuget packages on master

Releasing to nuget by tagging a commit will cause a github action to run that creates a proper release (release notes + api difference reports)

PR #85, opened 2020-06-03
Stand alone logger provider for Microsoft.Extensions.Logging that uses ECS to log directly to Elasticsearch

sgryphon:
Component donated from https://github.com/sgryphon/essential-logging
Re-licensed as Apache 2 (contributer agreement signed).
Once merged, I will deprecate from the other repository and refer here.
Also once merged I can update the pull request with the usage example to refer to the project directly.


src/Elastic.CommonSchema.LoggerProvider/build.ps1
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env pwsh
```
Mpdreamz, line 1: We can remove this file, will follow up making this project part of the existing build 👍

src/Elastic.CommonSchema.LoggerProvider/Elastic.CommonSchema.LoggerProvider.csproj
```diff
@@ -0,0 +1,38 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>netstandard2.0</TargetFramework>
+      <Product>Elastic.CommonSchema</Product>
+      <Authors>sgryphon</Authors>
+      <Copyright>Copyright (c) 2014-2020 by Elasticsearch BV</Copyright>
+      <Version>1.0.0+bd3ad6</Version>
+      <RootNamespace>Essential</RootNamespace>
```
Mpdreamz, line 9: Old root namespace

src/Elastic.CommonSchema.LoggerProvider/.config/dotnet-tools.json
```diff
@@ -0,0 +1,12 @@
+{
```
Mpdreamz, line 1: We can remove this file, our build takes care of versioning and with #80 we are moving to `minver.

src/Elastic.CommonSchema.LoggerProvider/Elastic.CommonSchema.LoggerProvider.csproj
```diff
@@ -0,0 +1,38 @@
...
+        <LangVersion>8</LangVersion>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+
+  <PropertyGroup>
+    <Description>Elasticsearch-Logstash-Kibana (ELK) stack logger provider for Microsoft.Extensions.Logging. Writes direct to Elasticsearch using the Elastic Common Schema (ECS), with semantic logging of structured data from message and scope values. The results can be viewed and queried in the Kibana console. 
+      
+See the project site on GitHub for examples and documentation.</Description>
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
```
Mpdreamz, line 18: Should be false

src/Elastic.CommonSchema.LoggerProvider/build.ps1
sgryphon: Removed.

src/Elastic.CommonSchema.LoggerProvider/.config/dotnet-tools.json
sgryphon: Removed (it shouldn't really be in a subfolder anyway). I generally use Gitversion; I saw some comments regarding it, although I've been using it with some Github actions and it works okay..

sgryphon:
> Did you want to port the test project as part of this PR or do you intend to follow up?

I don't have a unit test project (I did have for my Framework System.Diagnostics trace listener libraries, but haven't ported them across to Core Extensions.Logging ... it is on my todo list)

I'll do a separate PR for the example / how to use.

sgryphon:
The Jenkins build (Windows) had failed on the signing. I updated the project to use the shared properties which should fix this.

src/Elastic.CommonSchema.LoggerProvider/ConnectionPoolType.cs
```diff
@@ -0,0 +1,17 @@
+// Licensed to Elasticsearch B.V under one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+namespace Elastic.CommonSchema
```
Mpdreamz, line 5: The namespace for all files should be `Elastic.CommonSchema.LoggerProvider`

Mpdreamz: I wonder if `Elastic.CommonSchema.Extensions.Logging` would be a more descriptive name for this project. 


@russcam @gregkalapos thoughts?

Mpdreamz: I actually want `Elasticearch` in the name to signal it does live ingestion of logs to `Elasticsearch` but `Elastic.CommonSchema.Extension.Logging.Elasticsearch` reads horribly. Anyone willing to bikeshed this one ?

src/Elastic.CommonSchema.LoggerProvider/ElasticsearchDataProcessor.cs
```diff
@@ -0,0 +1,262 @@
...
+		{
+			var indexTime = logEvent.Timestamp ?? ElasticsearchLoggerProvider.LocalDateTimeProvider();
+			if (_options.IndexOffset.HasValue) indexTime = indexTime.ToOffset(_options.IndexOffset.Value);
+
+			var index = string.Format(_options.Index, indexTime);
+
+			var id = Guid.NewGuid().ToString();
+
+			var localClient = _lowLevelClient;
+			var response = localClient.Index<StringResponse>(index, id, PostData.Serializable(logEvent));
```
Mpdreamz, line 199: I think for the first version this is fine but we'd want to look into batching support via `Bulk()` for every N messages or N seconds do a flush. 

It'd be good to move from BlockingCollection to either TPL or channels for this: 

https://michaelscodingspot.com/performance-of-producer-consumer/

TPL is particulary strong in creating blocks out of streams.

src/Elastic.CommonSchema.LoggerProvider/ElasticsearchDataProcessor.cs
```diff
@@ -0,0 +1,262 @@
...
+					// case ConnectionPoolType.StickySniffing:
+					// case ConnectionPoolType.Cloud:
+					default:
+						throw new NotSupportedException($"Unknown connection pool type {_options.ConnectionPoolType}");
+				}
+
+				settings = new ConnectionConfiguration(connectionPool);
+			}
+
+			var lowlevelClient = new ElasticLowLevelClient(settings);
```
Mpdreamz, line 257:
```suggestion
			var lowlevelClient = new ElasticLowLevelClient(settings);
			((ElasticLowLevelClient)_lowLevelClient).Settings.Dispose();
```

Some background: https://github.com/elastic/elasticsearch-net/issues/4018

We initially did not want to advertise `IDisposable` on the client since we want folks to reuse instances agressively. We will make the client disposable with `8.0.0` most likely.

Mpdreamz: Reloading the client will affect in flight requests and could potentially throw abort exceptions. 

Is this something we really need to support or adding in because the reload trigger is there? 

We could potentially only support reloading new endpoints with:

```
((ElasticLowLevelClient)_lowLevelClient).Settings.ConnectionPool.Reseed(nodes);
```

Without destroying the client, other settings would require a restart though.

src/Elastic.CommonSchema.LoggerProvider/ConnectionPoolType.cs
sgryphon: Naming is important.

Probably don't want to repeat the class suffix in the namespace (also consider the 'using ...' statement needed for the LoggingBuilder extensions): Elastic.CommonSchema.LoggerProvider.ElasticsearchLoggerProvider 

I used Elastic.CommonSchema to be consistent with the other project components, and didn't want it too long, but the project/package name ended up inconsistent with the namespace.

If you want other suggestions, when searching for the library, people may use terms like "Elasticsearch LoggerProvider", and shorter is better:
* using Elastic.CommonSchema.Logging; & class Elastic.CommonSchema.Logging.ElasticsearchLoggerProvider
* using Elastic.Logging; & class Elastic.Logging.ElasticsearchLoggerProvider
* using Elastic.Extensions.Logging; & class Elastic.Extensions.Logging.ElasticsearchLoggerProvider

.NET uses Microsoft.Extensions.Logging.Console.ConsoleLoggerProvider, but the ConsoleLoggerExtensions only needs "using Microsoft.Extensions.Logging;", hence my initial use of just "using Elastic.CommonSchema:" at least for the builder.

src/Elastic.CommonSchema.LoggerProvider/ElasticsearchDataProcessor.cs
sgryphon: I have used batching in my previous libraries, but this is only a first version of this one.

One important feature for logging batching, from previous work, is that the first log message should send immediately (single message), then start batching. A user wants to see that first log message immediately, to know the system is working. You don't want to wait seconds / lots of messages before sending the first.

I've used channels before, and they are really good ... but I tried to keep it netstandard20 compliant.

src/Elastic.CommonSchema.LoggerProvider/ElasticsearchDataProcessor.cs
sgryphon: Not sure I understand your suggestion. If you are saying I should dispose of the client (or it's child), then I would do that after the atomic/interlocked swap, i.e. get rid of the old one.

PostEvent does grab a local copy, and configuration would not reload very often (if ever), so leaving one old one around for GC Disposable shouldn't be an issue. 

And, as you said, if you dispose of it while there is an inflight request that might cause an issue (which is why I create a new one, then do an atomic swap).

sgryphon: If only nodes changed, then it would be good to just reload them -- a good future enhancement.

sgryphon: If you have disposable fields, then the client should probably be disposable, even if intended to be a singleton. But you can also have a explicit implementation of the interface which makes it slightly less likely users will just be disposing you.

And I'm aware of the damned-if-you-do-damned-if-you-dont issues with either disposing or not disposing of HttpClientHandler... the improvements in 3.0/3.1 seem to handle it quite well with the single HttpClient / internal pool of connections, and having middleware for things like test injection, policy based retries, etc.

src/Elastic.CommonSchema.LoggerProvider/ElasticsearchDataProcessor.cs
sgryphon: Before batching though, I'd suggest the next needed feature I had on my todo list was supporting security/config options for the connection. (next feature)

src/Elastic.CommonSchema.LoggerProvider/ConnectionPoolType.cs
sgryphon: Happy to change this if you decide. Elastic.Extensions.Logging?

src/Elastic.CommonSchema.LoggerProvider/ElasticsearchDataProcessor.cs
sgryphon: This is for a future version; maybe move to an issue / project card to track?

src/Elastic.CommonSchema.LoggerProvider/ElasticsearchDataProcessor.cs
sgryphon: I suggest also move this to an issue, for a future change. (Although I'm not sure there is a better solution, especially given config changes would be rare.)

sgryphon:
This is approved, but I don't have permission to merge.

From the code review, the only open question is whether to change the namespace, e.g. Elastic.Extensions.Logging or Elasticsearch.Extensions.Logging (if you want Elasticsearch in there)?

sgryphon:
I decided to change the namespace to "Elasticsearch.Extensions.Logging", which has the Elasticsearch name in there (where it is writing to directly) and reflects it is for Extensions.Logging.

src/Elasticsearch.Extensions.Logging/ElasticsearchDataProcessor.cs
```diff
@@ -0,0 +1,263 @@
...
+		}
+
+		private void PostEvent(LogEvent logEvent)
+		{
+			var indexTime = logEvent.Timestamp ?? ElasticsearchLoggerProvider.LocalDateTimeProvider();
+			if (_options.IndexOffset.HasValue) indexTime = indexTime.ToOffset(_options.IndexOffset.Value);
+
+			var index = string.Format(_options.Index, indexTime);
+
+			var id = Guid.NewGuid().ToString();
```
russcam, line 197: If the id for a document is auto-generated, it'd be better to not send an id and rely on Elasticsearch to generate one. This is faster in Elasticsearch because it does not need to check if a document already exists with a given id.

src/Elasticsearch.Extensions.Logging/ElasticsearchDataProcessor.cs
```diff
@@ -0,0 +1,263 @@
...
+
+			var localClient = _lowLevelClient;
+			var response = localClient.Index<StringResponse>(index, id, PostData.Serializable(logEvent));
```
russcam, line 200: Should a failed API call be handled or logged?

src/Elasticsearch.Extensions.Logging/ElasticsearchDataProcessor.cs
```diff
@@ -0,0 +1,263 @@
...
+		}
+
+		private void ProcessLogQueue()
+		{
+			try
+			{
+				foreach (var logEvent in _messageQueue.GetConsumingEnumerable()) PostEvent(logEvent);
```
russcam, line 207: Using the bulk API to send a batch of events in one request would be preferable to sending individual log events.

src/Elasticsearch.Extensions.Logging/ElasticsearchLogger.cs
```diff
@@ -0,0 +1,359 @@
...
+
+					var formattedScope = isFormattedLogValues ? scope.ToString() : FormatValue(scope);
+					logEvent.Scopes.Add(formattedScope);
+				}, logEvent);
+			}
+		}
+
+		private void AddStateValues<TState>(TState state, LogEvent logEvent)
+		{
+			if (state is IEnumerable<KeyValuePair<string, object>> stateValues)
```
russcam, line 111: Is `state` always guaranteed to be an `IEnumerable<KeyValuePair<string, object>>`? Should something happen if it isn't e.g. log an error?

src/Elasticsearch.Extensions.Logging/ElasticsearchLogger.cs
```diff
@@ -0,0 +1,359 @@
...
+			if (activity != null)
+			{
+				// Unique identifier of the trace.
+				// A trace groups multiple events like transactions that belong together. For example, a user request handled by multiple inter-connected services.
+				logEvent.Trace = new Trace() { Id = activity.RootId };
+
+				// Unique identifier of the transaction.
+				// A transaction is the highest level of work measured within a service, such as a request to a server.
+				var spanId = ExtractW3cSpanIdFromActivityId(activity.Id);
+				logEvent.Transaction = new Transaction() { Id = spanId ?? activity.Id };
```
gregkalapos, line 146: I think setting `Transaction.Id` to `Activity.Current.Id` or to `SpanId` in case it's in W3C format is not correct. As the definition of `transaction` in ecs says (which is in the comment above) "transaction is the highest level of work measured within a service". The thinking behind this is that if you filter based on `transaction.id` you'll see all logs for the given transaction e.g. for a given incoming HTTP request (from the point when it hits the request pipeline until the response is sent back) in a single service. The scope of SpanId or `Activity.Id` is smaller - it'd be something like an outgoing HTTP request or a database call - but each `activity` will have its own id, so if we use those, then the filtering based on `Transaction.Id` will be very different, you'll only see logs for a given span (or activity), but not for the whole transaction.

In [Elastic APM we have the definition of a transaction](https://www.elastic.co/guide/en/apm/get-started/current/transactions.html) which is not the same as a span - `Activity` fits more the definition of the [span](https://www.elastic.co/guide/en/apm/get-started/current/transaction-spans.html). 

We have tracing integration with Serilog and NLog with the Elastic APM Agent (meaning we fill the 2 ecs tracing fields (`Trace.Id` and `Transaction.Id`) based on ids generated by the APM Agent) and in that case we set `Transaction.Id` to the real transaction id - which is the same for all spans within a single service.

Overall I'm afraid it'd be impossible to get a transaction.id just based on `Activity` here and I feel it'd be best to not fill `Transaction.Id` here. We could reiterate on this later, but for now I think it'd be better to just not set it to something which would be technically incorrect. So my suggestion is that we should remove setting the `logEvent.Transaction` part.

Happy to discuss this further, let me know if someone has different opinion.

src/Elasticsearch.Extensions.Logging/ElasticsearchLogger.cs
```diff
@@ -0,0 +1,359 @@
...
+
+				return true;
+			}
+
+			return false;
+		}
+
+		private string? ExtractW3cSpanIdFromActivityId(string activityId)
+		{
+			if (activityId.Length != 2 + 32 + 16 + 2 + 3) return null;
```
gregkalapos, line 234: Probably an edge case, but I think a non-W3C activityId can also be `2 + 32 + 16 + 2 + 3` long. 

An idea: if it's not W3C then it's probably the hierarchical id and that always starts with `|`, so a check like this would also catch that:

```
if (activityId[0] == '|' || activityId.Length != 2 + 32 + 16 + 2 + 3) return null;
```

Mpdreamz:
@russcam @gregkalapos discussed this over zoom and decided to merge this as is and follow up with smaller PR's to get this ready for release.

For now the new project is marked as non pack-able to prevent a release.

I will open a new ticket to keep track of the work needed.

Mpdreamz:
Thanks again for your hard work on this @sgryphon!

src/Elasticsearch.Extensions.Logging/Elasticsearch.Extensions.Logging.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <Title>Elasticsearch Logger Provider</Title>
+    <Description>Elasticsearch logger provider for Microsoft.Extensions.Logging. Writes direct to Elasticsearch using the Elastic Common Schema (ECS), with semantic logging of structured data from message and scope values, for use with the Elasticsearch-Logstash-Kibana (ELK) stack. The results can be viewed and queried in the Kibana console.</Description>
+    <PackageTags>Logging;LoggerProvider;Elasticsearch;ELK;Kibana;Logstash;Tracing;Diagnostics;Log;Trace;ECS</PackageTags>
+    <LangVersion>8</LangVersion>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Elasticsearch.Net" Version="7.6.1" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Configuration" Version="2.2.0" />
```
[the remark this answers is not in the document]
sgryphon: Yes, downgraded to 2.1

src/Elasticsearch.Extensions.Logging/ElasticsearchLoggerOptions.cs
```diff
@@ -0,0 +1,71 @@
...
+		/// <summary>
+		/// Gets or sets a flag indicating whether user details should be included in the message. Defaults to <c>true</c>.
+		/// </summary>
+		public bool IncludeUser { get; set; } = true;
+
+		/// <summary>
+		/// Gets or sets the format string for the Elastic search index. The current <c>DateTimeOffset</c> is passed as parameter
+		/// 0.
+		/// </summary>
+		public string Index { get; set; } = "dotnet-{0:yyyy.MM.dd}";
```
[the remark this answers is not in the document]
sgryphon: I don't know why other frameworks decided to use the same identifier as logstash. I know logstash is part of the ELK framework, but don't see any reason to masquerade as it. In fact it could be confusing if you also have logstash pushing events.  Maybe logstash is just that great that everyone wants to be them?

The value is configurable, so if there is some hard coded third party tool that is specifically looking for the logstash-* pattern, you can always change the index value.

src/Elasticsearch.Extensions.Logging/ElasticsearchLogger.cs
sgryphon: State could be anything, called from the public Log<TState>() method.  However usually it is not called directly, and either the high performance LoggerMessage.Define<>() is used, or one of the convenience extension methods, e.g. LogWarning().

The usual calls convert the passed in parameters to an internal state class, which is what you get access to in a logger provider. If you are just logging the message you can use the passed in formatter function.

However, for semantic logging you want to get out the individual parameter values, which I then put into the ECS labels dictionary. To do this we assume if the class implements IEnumerable<KeyValuePair<string,object>>, which the internal Microsoft class does, then we can use it to get the semantic labels and values.

There is a similar assumption made about BeginScope<TState>() to pull out additional semantic values. As we don't have access to the actual internal code, this is the best we can do and just handle it gracefully if the internals ever change. 

Nic Blumhardt blogged about it for Seq handling of Scope: https://nblumhardt.com/2016/11/ilogger-beginscope/

PR #90, opened 2020-06-19
Feature/serilog property filter

ghayes7:
Need a way to filter embedded Serilog LogEvent properties from being added to serialized / converted ECS Base.Metadata.

For example if I have embedded properties in my Serilog log event's that I used in the EcsTextFormatterConfiguration.MapCustom to enrich the ECS fields, I won't want it to be added to metadata. There's no other way to filter out a log event property, can't even remove it via EcsTextFormatterConfiguration.MapCustom because it's called after GetMetadata in LogEventConverter.ConvertToEcs. Even if the order was rearranged, still could not remove in MapCustom (i.e. LogEvent.RemovePropertyIfPresent) because the LogEvent is passed by reference so any subsequent renderings of a given logEvent, to different targets will not contain the previously removed property.

Also need a way to set the Serilog Tests LogTestBase classes Formatter property.


src/Elastic.CommonSchema.Serilog/EcsTextFormatterConfiguration.cs
```diff
@@ -19,6 +20,7 @@ public interface IEcsTextFormatterConfiguration
 		Func<Base, LogEvent, Base> MapCustom { get; set; }
 		bool MapExceptions { get; set; }
 		IHttpAdapter MapHttpAdapter { get; set; }
+		IList<string> LogEventPropertiesToFilter { get;set; }
```
russcam, line 23: This might be better as an `ISet<string>` 

1. for fast lookup
2. to allow String comparison implementation to be defined by the user as part of the concrete `ISet<string>` implementation

ghayes7: Excellent point! Updated with bbcb645502f950d8708911ea91aaed8ca19f06c7

Mpdreamz:
jenkins test this please

ghayes7:
@Mpdreamz @russcam Hey guys, anything I need to do to get this merged in and closed?

russcam:
@ghayes7, I've been out on PTO and looking at this again now

src/Elastic.CommonSchema.Serilog/EcsTextFormatterConfiguration.cs
```diff
@@ -27,6 +29,7 @@ public class EcsTextFormatterConfiguration : IEcsTextFormatterConfiguration
 		bool IEcsTextFormatterConfiguration.MapCurrentThread { get; set; } = true;
 
 		IHttpAdapter IEcsTextFormatterConfiguration.MapHttpAdapter { get; set; }
+		public ISet<string> LogEventPropertiesToFilter { get; set; }
```
russcam, line 32: nitpick: I think this should be explicitly implemented in line with the other properties, with a fluent method to set, to keep the API consistent.

src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
```diff
@@ -91,7 +91,7 @@ private static Trace GetTrace(LogEvent logEvent) => !logEvent.TryGetScalarProper
 				? null
 				: new Transaction { Id = transactionId.Value.ToString() };
 
-		private static IDictionary<string, object> GetMetadata(LogEvent logEvent)
+		private static IDictionary<string, object> GetMetadata(LogEvent logEvent, ISet<string> propKeyFilter=null)
```
russcam, line 94:
```suggestion
		private static IDictionary<string, object> GetMetadata(LogEvent logEvent, ISet<string> logEventPropertiesToFilter)
```

Don't think this needs the `null` default value

src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
```diff
@@ -137,7 +137,9 @@ private static Trace GetTrace(LogEvent logEvent) => !logEvent.TryGetScalarProper
 					case SpecialKeys.MachineName:
 						continue;
 				}
-
+				//key present in list of keys to filter
+				if (propKeyFilter?.Contains(logEventPropertyValue.Key) ?? false)
```
russcam, line 141:
```suggestion
				if (logEventPropertiesToFilter?.Contains(logEventPropertyValue.Key) ?? false)
```

src/Elastic.CommonSchema.Serilog/EcsTextFormatterConfiguration.cs
ghayes7: Hey, thanks for the review! I don't think I'm following your line of thought exactly, could you clarify?
Do you mean like:
```suggestion
		public ISet<string> LogEventPropertiesToFilter { get; set; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
		public EcsTextFormatterConfiguration FilterLogEventProperties(ISet<string> value) => Assign(this, value, (o, v) => o.LogEventPropertiesToFilter = v);
```
Would still want it to be public to allow manual updates (add, remove, ContainsKey, etc), right? Though I guess technically, you could do updates through the FilterLogEventProperties fluent setter (albeit quite inefficiently)? 

Or am I misunderstanding?

russcam: No worries! I think the interface property should be explicitly implemented with a fluent method, in line with other properties e.g.

```c#
ISet<string> IEcsTextFormatterConfiguration.LogEventPropertiesToFilter { get; set; }

public EcsTextFormatterConfiguration LogEventPropertiesToFilter(ISet<string> value) => Assign(this, value, (o, v) => o.LogEventPropertiesToFilter = v);
```

I don't know the reasoning why these are fluent methods (a style choice?), but I guess typical usage would be to set once, for the lifetime of the running app, since the configuration is used by `EcsTextFormatter`.

>  Though I guess technically, you could do updates through the FilterLogEventProperties fluent setter (albeit quite inefficiently)?

You would also be able to make changes to the `ISet<string>` assigned by casting to the `IEcsTextFormatterConfiguration` interface.

ghayes7: @russcam Done, see https://github.com/elastic/ecs-dotnet/pull/90/commits/8eceb1b2e0518557da6509e01baa6634249c8691

ghayes7:
@russcam I think I got everything. Though, github's still saying I missed a suggestion, am I blind?

russcam:
Thanks for your contribution @ghayes7

Issue #95, opened 2020-07-14
[meta] Elasticsearch.Extensions.Logging GA


Mpdreamz:
This ticket is a continuation of the following PR: https://github.com/elastic/ecs-dotnet/pull/85

- [x] Rely on Elasticsearch to generate the id
- [x] Move to a simple buffering approach and `_bulk`
- [x] Add a hook to be notified of failures
- [x] [`state` always an ` IEnumerable<KeyValuePair<string, object>>`](https://github.com/elastic/ecs-dotnet/pull/85#discussion_r441971235)?
- [x] Make sure `trace.id` flows into Elasticsearch so that Elastic APM can corrolate logs. (done https://github.com/elastic/ecs-dotnet/pull/96 )
- [x] Downgrade `Microsoft.Extensions.*` packages to `2.1.0`.
- [x] Should we default to `logstash-[current_date]` like the serilog sink does?
- [x] Remove `.Cast<char>` 
- [x] Remove unused `_w3CFormat` regular expression
- [x] `ExtractW3cSpanIdFromActivityId` has a possible detection edgecase that needs to be adressed. (https://github.com/elastic/ecs-dotnet/pull/96 eliminated it)
- [x] Introduce integration tests for this project.
- [x] Add support for `CloudConnectionPool` (https://github.com/elastic/ecs-dotnet/pull/98) ty @btastic !
- [x] [Make sure CreateConnectionPool throws instead of returning `null`](https://github.com/elastic/ecs-dotnet/pull/97#discussion_r457865292)

When all that is done we can:

- [x] Remove `<IsPackable>false</IsPackable>` from the project 

That should be enough to create and publish the package when we tag the next version.

cc @russcam @gregkalapos @sgryphon feel free to take an item and submit PR's. If it's a bigger chunk of work leave a note so we don't duplicate efforts.

btastic:
Hello,

I just found this package while trying the elastic cloud trial. However, this logger does not seem to have the capability connect to the elastic cloud with its current API. Are there any plans to add that?

I'd like to contribute if you think this is a good addition.

Questions:
- As far as I can tell from the source code, you can add multiple Nodes to one configuration. That will make it not straight forward to define different ConnectionPoolTypes for each node. If the ConnectionPoolType is Cloud, you'd have to define it differently than other nodes as far as I understand. Or am I missing something?

Mpdreamz:
Hey @btastic I think its a fantastic addition, looking forward to your PR 😄 

[...]

PR #97, opened 2020-07-16
Several improvements to Elasticsearch.Ext.Logging

Mpdreamz:
- Add some basic integration tests to assert the logs make it in the right index
- Rename `ElasticsearchDataProcessor` to `ElasticsearchDataShipper`
- Move enrichtment of `LogEvent` out of the shipper and into static helper class.
- Rely on elasticsearch to generate ids
- Move to PostData.StreamHandler to serialize directly to the IO stream.

relates: https://github.com/elastic/ecs-dotnet/issues/95


src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
```diff
@@ -0,0 +1,135 @@
...
+		}
+
+		public void Dispose()
+		{
+			_messageQueue?.CompleteAdding();
+			try
+			{
+				_outputThread.Join(1500); // with timeout in case writer is locked
+			}
+			catch (ThreadStateException) { }
```
russcam, line 50: is there an _internal_ logger to which this exception can be written?

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
```diff
@@ -0,0 +1,135 @@
...
+			{
+				// This is SingleNode with "http://localhost:9200"
+				settings = new ConnectionConfiguration();
+			}
+			else if (_options.ConnectionPoolType == ConnectionPoolType.SingleNode
+				|| _options.ConnectionPoolType == ConnectionPoolType.Unknown && nodes.Length == 1)
+				settings = new ConnectionConfiguration(nodes[0]);
+			else
+			{
+				settings = new ConnectionConfiguration(connectionPool);
```
russcam, line 127: Looks like `_options.ShipTo.CreateConnectionPool();` can return `null`: 

https://github.com/elastic/ecs-dotnet/blob/249b86d9163bd373182ef5d458910a98f7a7973e/src/Elasticsearch.Extensions.Logging/ShipTo.cs#L61-L86

but `ConnectionConfiguration` does not guard against `null`: 

https://github.com/elastic/elasticsearch-net/blob/f1ffbf260132eb83ea4e7950c1bcf19ef679ee6f/src/Elasticsearch.Net/Configuration/ConnectionConfiguration.cs#L210

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
```diff
@@ -0,0 +1,135 @@
...
+			}
+
+			var lowlevelClient = new ElasticLowLevelClient(settings);
+
+			_ = Interlocked.Exchange(ref _lowLevelClient, lowlevelClient);
```
russcam, line 132: I'm struggling to see what benefit this provides over assigning to the field directly in this case, as the original value is not used and field assignment is atomic on all .NET platforms AFAIK.  Is there a benefit that I'm missing?

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
```diff
@@ -0,0 +1,135 @@
...
+				foreach (var logEvent in _messageQueue.GetConsumingEnumerable()) PostEvent(logEvent);
+			}
+			catch
+			{
+				try
+				{
+					_messageQueue.CompleteAdding();
+				}
+				catch { }
+			}
```
russcam, line 105: If I understand this correctly, an exception in posting the event will mark the blocking collection as no longer accepting items, causing any subsequent items to be sent to `PostEvent` directly in `EnqueueMessage`. Would we want to distinguish the types of exceptions that should cause this behaviour vs. those that shouldn't e.g. are there exceptions the client could throw where this behaviour is undesirable?

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
```diff
@@ -0,0 +1,135 @@
...
+
+		private void PostEvent(LogEvent logEvent)
+		{
+			var indexTime = logEvent.Timestamp ?? ElasticsearchLoggerProvider.LocalDateTimeProvider();
+			if (_options.IndexOffset.HasValue) indexTime = indexTime.ToOffset(_options.IndexOffset.Value);
+
+			var index = string.Format(_options.Index, indexTime);
+
+			var localClient = _lowLevelClient;
+			var response = localClient.Index<StringResponse>(index,
```
russcam, line 83: I had a question about bulk, but see that this is addressed in #99 😄

tests/Elasticsearch.Extensions.Logging.IntegrationTests/LoggingTests.cs
```diff
@@ -0,0 +1,75 @@
...
+
+		[Fact]
+		public async Task LogsEndUpInCluster()
+		{
+			using var _ = CreateLogger(out var logger, out var indexPrefix);
+
+			logger.LogError("an error occured");
+
+			// TODO make sure we can await something here on ElasticsearchDataShipper
+			await Task.Delay(TimeSpan.FromSeconds(5));
```
russcam, line 62: Could the logger expose a property/event to ascertain that it has sent or is sending events?

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
Mpdreamz: I looked into fallback loggers in the Mic.Ext.Logging world but there is no such concept. The best we can do is expose callbacks to allow to user to log manually to fallback locations in their application configuration.

This is a carry over of #85 (introduced there not here) that is rewritten in #98.

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
Mpdreamz: Good catch!, this is a carry over from #98, will update the meta ticket #95  to follow up in a separate PR

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
Mpdreamz: This is a carry over from #85 and is now rewritten to move away from `BlockingCollection` to `Channels` in #99

tests/Elasticsearch.Extensions.Logging.IntegrationTests/LoggingTests.cs
Mpdreamz: That was my intend but I knew the shipper had to be rewritten so I left the TODO :) 

Once #99 lands we can expose more callbacks and stats

Mpdreamz:
Thanks for the review @russcam I think most comments should be addressed in #99 

Noted the need for `CreateConnectionPool` not to return null in the meta ticket so we can address this in a smaller dedicated PR.

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
Mpdreamz: There is not, carry over from #85 we can address that in a follow up PR.

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
```diff
@@ -0,0 +1,135 @@
+// Licensed to Elasticsearch B.V under one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+using System;
+using System.Collections.Concurrent;
+using System.Linq;
+using System.Threading;
+using Elastic.CommonSchema;
```
gregkalapos, line 9:
```suggestion
```

Mpdreamz: Will tackle later 👍

PR #69, opened 2020-04-03
Feature/console example elasticsearchloggerprovider

sgryphon:
Add an example that uses Essential.LoggerProvider.Elasticsearch, a stand alone logger provider that uses ECS. (i.e. not part of a third party logging framework like Serilog or NLog that goes through a bridge/adapter, but a logger provider that connects directly to Microsoft.Extensions.Logging).

The example also includes a docker compose file to get Elasticsearch + Kibana up and running locally, and also demonstrates the use of the Microsoft LoggerMessage for high performance logging.

** NOTE: I am the author of the ElasticsearchLoggerProvider package. **

PR #99, opened 2020-07-20
Bulk support for extensions logging integration.

Mpdreamz:
Continuation of #97

Introduces batching of log events before indexing into Elasticsearch. Powered by `System.Threading.Channels`. It supports sending every N items or every M interval after the first item was added. 

This also introduces the ability to have multiple consumers drain the channel concurrently. 

A callback now exists when items are being dropped because the channel is full and when a `_bulk` request has occurred. 

We still need to add 

- _bulk failure handling
- _bulk retry handling (exponential backoff)
- _bulk rejection callback for `LogEvent` 

Since this PR and #97 are big enough already will tackle the remained when we pull in the flight PR's. 

This also includes a modified version of the example project from @sgryphon 's PR: #69 

It also includes a high volume worker that can be started with `dotnet run -- high`.

### DIFF
--- a/examples/Elasticsearch.Extensions.Logging.Example/HighVolumeWorkSimulation.cs
+++ b/examples/Elasticsearch.Extensions.Logging.Example/HighVolumeWorkSimulation.cs
@@ -14,9 +14,9 @@ namespace Elasticsearch.Extensions.Logging.Example
 	/// <summary> Simulate work that logs in low volume with some time in between each log call </summary>
 	public class HighVolumeWorkSimulation : BackgroundService
 	{
-		private readonly ILogger<LowVolumeWorkSimulation> _logger;
+		private readonly ILogger<HighVolumeWorkSimulation> _logger;
 
-		public HighVolumeWorkSimulation(ILogger<LowVolumeWorkSimulation> logger) => _logger = logger;
+		public HighVolumeWorkSimulation(ILogger<HighVolumeWorkSimulation> logger) => _logger = logger;
 
 		protected override async Task ExecuteAsync(CancellationToken ctx)
 		{
--- a/examples/Elasticsearch.Extensions.Logging.Example/HighVolumeWorkSimulation.cs
+++ b/examples/Elasticsearch.Extensions.Logging.Example/HighVolumeWorkSimulation.cs
@@ -23,7 +23,7 @@ namespace Elasticsearch.Extensions.Logging.Example
 			for (var i = 0; i < 100_000; i++)
 			{
 				_logger.LogWarning($"We are logging way too much: {i}");
-				if (i % 1_00 == 0) await Task.Delay(1, ctx);
+				if (i % 100 == 0) await Task.Delay(1, ctx);
 			}
 		}
 	}
--- a/examples/Elasticsearch.Extensions.Logging.Example/HighVolumeWorkSimulation.cs
+++ b/examples/Elasticsearch.Extensions.Logging.Example/HighVolumeWorkSimulation.cs
@@ -1,0 +1,30 @@
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Security.Claims;
+using System.Security.Principal;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+namespace Elasticsearch.Extensions.Logging.Example
+{
+
+	/// <summary> Simulate work that logs in low volume with some time in between each log call </summary>
+	public class HighVolumeWorkSimulation : BackgroundService
+	{
+		private readonly ILogger<LowVolumeWorkSimulation> _logger;
+
+		public HighVolumeWorkSimulation(ILogger<LowVolumeWorkSimulation> logger) => _logger = logger;
+
+		protected override async Task ExecuteAsync(CancellationToken ctx)
+		{
+			for (var i = 0; i < 100_000; i++)
+			{
+				_logger.LogWarning($"We are logging way too much: {i}");
+				if (i % 1_00 == 0) await Task.Delay(1, ctx);
+			}
+		}
+	}
+}
--- a/src/Elasticsearch.Extensions.Logging/ConsumerBuffer.cs
+++ b/src/Elasticsearch.Extensions.Logging/ConsumerBuffer.cs
@@ -26,6 +26,7 @@ namespace Elasticsearch.Extensions.Logging
 
 		public TimeSpan ForceFlushAfter { get; }
 		public List<LogEvent> Buffer { get; }
+                /// <summary>The time that the first event is read from the channel and added to the buffer, from first read or after the buffer is reset.</summary>
 		private DateTimeOffset? TimeOfFirstRead { get; set; }
 
 		public int Count => Buffer.Count;
--- a/src/Elasticsearch.Extensions.Logging/ConsumerBuffer.cs
+++ b/src/Elasticsearch.Extensions.Logging/ConsumerBuffer.cs
@@ -1,0 +1,106 @@
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+
+namespace Elasticsearch.Extensions.Logging
+{
+	/// <summary>
+	/// Information about the consumers local buffer, available to users in <see cref="DrainThrottles.ElasticsearchResponseCallback" />
+	/// </summary>
+	public interface IConsumerBuffer
+	{
+		int Count { get; }
+		TimeSpan? DurationSinceFirstRead { get; }
+	}
+
+	/// <summary>
+	/// Acts as a wrapper around the consumers buffer of messages. We want the buffer to flush
+	/// every N messages OR in the case messages do no flow fast enough or stop before N messages were received every
+	/// M timespan.
+	/// </summary>
+	internal class ConsumerBuffer : IConsumerBuffer, IDisposable
+	{
+		private readonly int _maxBufferSize;
+
+		public TimeSpan ForceFlushAfter { get; }
+		public List<LogEvent> Buffer { get; }
+		private DateTimeOffset? TimeOfFirstRead { get; set; }
+
+		public int Count => Buffer.Count;
+		public TimeSpan? DurationSinceFirstRead => DateTimeOffset.UtcNow - TimeOfFirstRead;
+		public bool NoThresholdsHit => Count == 0 || (Count < _maxBufferSize && DurationSinceFirstRead <= ForceFlushAfter);
+
+		public ConsumerBuffer(int maxBufferSize, TimeSpan forceFlushAfter)
+		{
+			_maxBufferSize = maxBufferSize;
+			ForceFlushAfter = forceFlushAfter;
+			Buffer = new List<LogEvent>(maxBufferSize);
+			TimeOfFirstRead = null;
+		}
+
+		public void Add(LogEvent item)
+		{
+			TimeOfFirstRead ??= DateTimeOffset.UtcNow;
+			Buffer.Add(item);
+		}
+
+		public void Reset()
+		{
+			Buffer.Clear();
+			TimeOfFirstRead = null;
+		}
+
+		private TimeSpan Wait
+		{
+			get
+			{
+				if (!DurationSinceFirstRead.HasValue) return ForceFlushAfter;
+				var d = DurationSinceFirstRead.Value;
+				return d < ForceFlushAfter ? ForceFlushAfter - d : ForceFlushAfter;
+			}
+		}
+
+		private CancellationTokenSource _breaker = new CancellationTokenSource();
+
+
+		/// <summary>
+		/// Call <see cref="ChannelReader{T}.WaitToReadAsync"/> with a timeout to force a flush to happen every
+		/// <see cref="ForceFlushAfter"/>. This tries to avoid allocation too many <see cref="CancellationTokenSource"/>'s
+		/// needlessly and reuses them if possible. If we know the buffer is empty we can wait indefinitely as well.
+		/// </summary>
+		public async Task<bool> WaitToReadAsync(ChannelReader<LogEvent> reader)
+		{
+			//if we have nothing in the buffer wait indefinitely for messages
+			if (_breaker.IsCancellationRequested)
+			{
+				_breaker.Dispose();
+				_breaker = new CancellationTokenSource();
+			}
+
+			try
+			{
+				//if we have nothing in the buffer wait indefinitely for messages
+				var w = Count == 0 ? TimeSpan.FromMilliseconds(-1) : Wait;
+				_breaker.CancelAfter(w);
+				var _ = await reader.WaitToReadAsync(_breaker.Token).ConfigureAwait(false);
+				_breaker.CancelAfter(ForceFlushAfter);
+				return true;
+			}
+			catch (Exception) when (_breaker.IsCancellationRequested)
+			{
+				_breaker.Dispose();
+				_breaker = new CancellationTokenSource();
+				return true;
+			}
+			catch (Exception)
+			{
+				_breaker.CancelAfter(ForceFlushAfter);
+				return true;
+			}
+		}
+
+		public void Dispose() => _breaker.Dispose();
+	}
+}
--- a/src/Elasticsearch.Extensions.Logging/ConsumerBuffer.cs
+++ b/src/Elasticsearch.Extensions.Logging/ConsumerBuffer.cs
@@ -73,7 +73,6 @@ namespace Elasticsearch.Extensions.Logging
 		/// </summary>
 		public async Task<bool> WaitToReadAsync(ChannelReader<LogEvent> reader)
 		{
-			//if we have nothing in the buffer wait indefinitely for messages
 			if (_breaker.IsCancellationRequested)
 			{
 				_breaker.Dispose();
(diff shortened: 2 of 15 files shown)


examples/Elasticsearch.Extensions.Logging.Example/HighVolumeWorkSimulation.cs
```diff
@@ -0,0 +1,30 @@
...
+		private readonly ILogger<LowVolumeWorkSimulation> _logger;
+
+		public HighVolumeWorkSimulation(ILogger<LowVolumeWorkSimulation> logger) => _logger = logger;
+
+		protected override async Task ExecuteAsync(CancellationToken ctx)
+		{
+			for (var i = 0; i < 100_000; i++)
+			{
+				_logger.LogWarning($"We are logging way too much: {i}");
+				if (i % 1_00 == 0) await Task.Delay(1, ctx);
```
Mpdreamz, line 26:
```suggestion
				if (i % 100 == 0) await Task.Delay(1, ctx);
```

`1_00` is a weird way to format `100`, left over from when I wrote `1_000` 😄

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
```diff
@@ -3,106 +3,126 @@
...
+							},
+							async (@event, stream, ctx) =>
+							{
+								foreach (var logEvent in buffer.Buffer)
+								{
+									var indexTime = logEvent.Timestamp ?? ElasticsearchLoggerProvider.LocalDateTimeProvider();
+									if (_options.IndexOffset.HasValue) indexTime = indexTime.ToOffset(_options.IndexOffset.Value);
+
+									var index = string.Format(_options.Index, indexTime);
+									var indexHeader = new { index = new { _index = index } };
```
russcam, line 84: would there be some benefit to introducing a type for this?

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
```diff
@@ -3,106 +3,126 @@
...
 
-		public void Dispose()
-		{
-			_messageQueue?.CompleteAdding();
-			try
-			{
-				_outputThread.Join(1500); // with timeout in case writer is locked
-			}
-			catch (ThreadStateException) { }
+		private static readonly byte[] LineFeed = { 10 };
```
russcam, line 52:
```suggestion
		private static readonly byte[] LineFeed = { (byte)'\n' };
```

nitpick: perhaps more readable?

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
```diff
@@ -126,7 +146,7 @@ private void UpdateClient()
 			{
 				settings = new ConnectionConfiguration(connectionPool);
 			}
-
+			settings = settings.EnableDebugMode();
```
russcam, line 149: Think this shouldn't be included?

src/Elasticsearch.Extensions.Logging/ElasticsearchLoggerOptions.cs
```diff
@@ -68,6 +69,45 @@ public class ElasticsearchLoggerOptions
...
+
+	}
+
+	/// <summary>
+	/// Controls how <see cref="LogEvent"/>'s are batched and send to Elasticsearch. These can not be dynamically updated.
+	/// </summary>
+	public class DrainThrottles
+	{
+		/// <summary>
+		/// The maximum number of `<see cref="LogEvent"/> that can queued in memory. If this threshold is reached messages will be dropped
```
russcam, line 83:
```suggestion
		/// The maximum number of <see cref="LogEvent"/> that can be queued in memory. If this threshold is reached, events will be dropped
```

src/Elasticsearch.Extensions.Logging/ElasticsearchLoggerOptions.cs
```diff
@@ -68,6 +69,45 @@ public class ElasticsearchLoggerOptions
...
+		/// </summary>
+		public int MaxInFlightMessages { get; set; } = 100_000;
+
+		/// <summary>
+		/// The number of messages a local buffer should reach before sending the messages in a single call to `Elasticsearch`.
```
russcam, line 88:
```suggestion
		/// The number of events a local buffer should reach before sending the events in a single call to Elasticsearch.
```

src/Elasticsearch.Extensions.Logging/ElasticsearchLoggerOptions.cs
```diff
@@ -68,6 +69,45 @@ public class ElasticsearchLoggerOptions
...
+		/// </summary>
+		public int MaxConsumerBufferSize { get; set; } = 1_000;
+
+		/// <summary>
+		/// A consumer builds up a local buffer until <see cref="MaxConsumerBufferSize"/> is reached. If messages come in too slow however these
```
russcam, line 93:
```suggestion
		/// A consumer builds up a local buffer until <see cref="MaxConsumerBufferSize"/> is reached. If events come in too slow, these
```

src/Elasticsearch.Extensions.Logging/ElasticsearchLoggerOptions.cs
```diff
@@ -68,6 +69,45 @@ public class ElasticsearchLoggerOptions
...
+		/// messages could end up taking forever to be sent to Elasticsearch. This controls how long a buffer may exists before a flush is triggered.
```
russcam, line 94:
```suggestion
		/// events could end up taking forever to be sent to Elasticsearch. This controls how long a buffer may exist before a flush is triggered.
```

src/Elasticsearch.Extensions.Logging/ElasticsearchLoggerOptions.cs
```diff
@@ -68,6 +69,45 @@ public class ElasticsearchLoggerOptions
...
+		/// </summary>
+		public TimeSpan MaxConsumerBufferLifeTime { get; set; } = TimeSpan.FromSeconds(5);
+
+		/// <summary>
+		/// The maximum number of consumers allowed to poll for new messages on the channel. Defaults to 1, increase to introduce concurrency.
```
russcam, line 99:
```suggestion
		/// The maximum number of consumers allowed to poll for new events on the channel. Defaults to 1, increase to introduce concurrency.
```

src/Elasticsearch.Extensions.Logging/ElasticsearchLoggerOptions.cs
```diff
@@ -68,6 +69,45 @@ public class ElasticsearchLoggerOptions
...
+		/// </summary>
+		public int ConcurrentConsumers { get; set; } = 1;
+
+		/// <summary>
+		/// If <see cref="MaxInFlightMessages"/> is reached, <see cref="LogEvent"/>'s will fail to be published. You can be notified of dropped
+		/// messages with this callback
```
russcam, line 105:
```suggestion
		/// events with this callback
```

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
```diff
@@ -3,106 +3,126 @@
...
+			var maxConsumers = Math.Max(1, options.Throttles.ConcurrentConsumers);
+			PubSub = Channel.CreateBounded<LogEvent>(new BoundedChannelOptions(options.Throttles.MaxInFlightMessages)
+			{
+				SingleReader = maxConsumers == 1,
+				AllowSynchronousContinuations = true,
+				// wait does not block it simply signals that Writer.TryWrite should return false and be retried
+				// DropWrite will make `TryWrite` always return true, which is not what we want.
+				FullMode = BoundedChannelFullMode.Wait
+			});
+			async Task ConsumeMessages() =>
```
russcam, line 33: I think `ConsumeMessages` could be removed and `Consume` called directly in `_backgroundTasks.Add` taking `Func<Task>` on line 36.

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
```diff
@@ -3,106 +3,126 @@
...
+			for (var i = 0; i < maxConsumers; i++)
+				_backgroundTasks.Add(Task.Factory.StartNew(() => ConsumeMessages(), TaskCreationOptions.LongRunning));
 		}
 
+
 		private IElasticLowLevelClient _lowLevelClient = default!;
 
-		private ElasticsearchLoggerOptions _options = default!;
+		private ElasticsearchLoggerOptions _options;
+		public Channel<LogEvent> PubSub { get; }
```
russcam, line 43: minor nitpick: Maybe `Channel`?

src/Elasticsearch.Extensions.Logging/ConsumerBuffer.cs
```diff
@@ -0,0 +1,106 @@
...
+
+
+		/// <summary>
+		/// Call <see cref="ChannelReader{T}.WaitToReadAsync"/> with a timeout to force a flush to happen every
+		/// <see cref="ForceFlushAfter"/>. This tries to avoid allocation too many <see cref="CancellationTokenSource"/>'s
+		/// needlessly and reuses them if possible. If we know the buffer is empty we can wait indefinitely as well.
+		/// </summary>
+		public async Task<bool> WaitToReadAsync(ChannelReader<LogEvent> reader)
+		{
+			//if we have nothing in the buffer wait indefinitely for messages
```
russcam, line 75: looks like this comment is a copy/paste from below and should be removed?

src/Elasticsearch.Extensions.Logging/ElasticsearchLoggerOptions.cs
```diff
@@ -68,6 +69,45 @@ public class ElasticsearchLoggerOptions
...
+		/// <summary>
+		/// The number of messages a local buffer should reach before sending the messages in a single call to `Elasticsearch`.
+		/// </summary>
+		public int MaxConsumerBufferSize { get; set; } = 1_000;
+
+		/// <summary>
+		/// A consumer builds up a local buffer until <see cref="MaxConsumerBufferSize"/> is reached. If messages come in too slow however these
+		/// messages could end up taking forever to be sent to Elasticsearch. This controls how long a buffer may exists before a flush is triggered.
+		/// </summary>
+		public TimeSpan MaxConsumerBufferLifeTime { get; set; } = TimeSpan.FromSeconds(5);
```
russcam, line 96:
```suggestion
		public TimeSpan MaxConsumerBufferLifetime { get; set; } = TimeSpan.FromSeconds(5);
```

src/Elasticsearch.Extensions.Logging/ElasticsearchLoggerOptions.cs
```diff
@@ -68,6 +69,45 @@ public class ElasticsearchLoggerOptions
...
+
+		/// <summary>
+		/// The maximum number of consumers allowed to poll for new messages on the channel. Defaults to 1, increase to introduce concurrency.
+		/// </summary>
+		public int ConcurrentConsumers { get; set; } = 1;
+
+		/// <summary>
+		/// If <see cref="MaxInFlightMessages"/> is reached, <see cref="LogEvent"/>'s will fail to be published. You can be notified of dropped
```
russcam, line 104:
```suggestion
		/// If <see cref="MaxInFlightMessages"/> is reached, <see cref="LogEvent"/>'s will fail to be published to the channel. You can be notified of dropped
```

src/Elasticsearch.Extensions.Logging/ConsumerBuffer.cs
```diff
@@ -0,0 +1,106 @@
...
+	/// every N messages OR in the case messages do no flow fast enough or stop before N messages were received every
+	/// M timespan.
+	/// </summary>
+	internal class ConsumerBuffer : IConsumerBuffer, IDisposable
+	{
+		private readonly int _maxBufferSize;
+
+		public TimeSpan ForceFlushAfter { get; }
+		public List<LogEvent> Buffer { get; }
+		private DateTimeOffset? TimeOfFirstRead { get; set; }
```
russcam, line 29:
```suggestion
                /// <summary>The time that the first event is read from the channel and added to the buffer, from first read or after the buffer is reset.</summary>
		private DateTimeOffset? TimeOfFirstRead { get; set; }
```

src/Elasticsearch.Extensions.Logging/ConsumerBuffer.cs
```diff
@@ -0,0 +1,106 @@
...
+				_breaker.Dispose();
+				_breaker = new CancellationTokenSource();
+				return true;
+			}
+			catch (Exception)
+			{
+				_breaker.CancelAfter(ForceFlushAfter);
+				return true;
+			}
+		}
```
russcam, line 102: This looks like using exceptions for control flow. Is there an approach that doesn't use try/catch?

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
Mpdreamz: 👍 absolutely

src/Elasticsearch.Extensions.Logging/ConsumerBuffer.cs
Mpdreamz: It is but i see no way around it,  the cancellation will always throw + the try catch is there also to never crash the user application. 

I do try to minimize the throwing by waiting indefinitely if we know there is no work to be done. In a high load use case the `_breaker` is less likely to hit the timeout as well.

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
Mpdreamz: ++ left over from debugging

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
Mpdreamz: ++, I could have sworn I renamed already.

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
Mpdreamz: There would! Was holding off on dedicated types when following this up with retries, backoff and response failure callbacks.

I'm in two minds to spin off `ElasticsearchDataShipper` as its own package which would benefit from some more typing and control for the end user as well.

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
Mpdreamz: It should not!

src/Elasticsearch.Extensions.Logging/ElasticsearchDataShipper.cs
russcam: I was thinking along the lines of

```c#
private readonly List<Task> _backgroundTasks =  new List<Task>();
```

```c#
for (var i = 0; i < maxConsumers; i++)
    _backgroundTasks.Add(Task.Factory.StartNew(() => Consume(options), TaskCreationOptions.LongRunning));
```

with `Consume` taking `ElasticsearchLoggerOptions` so that changes to options will affect consumers _(might be unexpected if changes to `options.Throttles` don't affect existing consumers)_.

src/Elasticsearch.Extensions.Logging/ElasticsearchLoggerOptions.cs
```diff
@@ -68,6 +69,45 @@ public class ElasticsearchLoggerOptions
...
 		public string[] Tags { get; set; } = new string[0];
+
+		public DrainThrottles Throttles { get; set; } = new DrainThrottles();
+
+	}
+
+	/// <summary>
+	/// Controls how <see cref="LogEvent"/>'s are batched and send to Elasticsearch. These can not be dynamically updated.
+	/// </summary>
+	public class DrainThrottles
```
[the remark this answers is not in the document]
Mpdreamz: Went with `BufferOptions`, was trying to be more imaginative then an Options object under another Options object but `DrainThrottles` feels like a stretch indeed.

src/Elasticsearch.Extensions.Logging/ElasticsearchLoggerOptions.cs
```diff
@@ -68,6 +69,45 @@ public class ElasticsearchLoggerOptions
...
+		/// </summary>
+		public int ConcurrentConsumers { get; set; } = 1;
+
+		/// <summary>
+		/// If <see cref="MaxInFlightMessages"/> is reached, <see cref="LogEvent"/>'s will fail to be published to the channel. You can be notified of dropped
+		/// events with this callback
+		/// </summary>
+		public Action<LogEvent> PublishRejectionCallback { get; set; } = e => { };
+
+		public Action<IElasticsearchResponse, IConsumerBuffer> ElasticsearchResponseCallback { get; set; } = (r, b) => { };
```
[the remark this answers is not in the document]
Mpdreamz: This makes sense since you might want to subscribe multiple times in different layers of your application. At the very least we don't know if folks want to always ONLY have 1 listener.

The callbacks are not used for control flow either. Added a TODO to follow up on in a dedicated PR.

Mpdreamz:
@russcam I think this is ready for another round. Your comment about `throttles` and consume task options is a good one but the `throttle` options are documented as not picking changes. In general I would like to minimizing support live config changes as much as we need to. 

I don't want to support scaling consumers and buffer sizes just yet, most likely worth their own PR 😄

PR #100, opened 2020-08-12
Move actual shipper into own project

Mpdreamz:
This moves the shipping component from `Elasticsearch.Extensions.Logging` into its own package.

This allows any application to new an `ElasticsearchChannel<TEvent>` and
use it to push data into through  `TryWrite()` . The channel will then take care of
buffering and retrying etc.


Mpdreamz:
Merging this in since we previously discussed this move. @russcam @gregkalapos I am not sold on `Elastic.Ingest` completely anyone willing to 🚲-shed this one? 😸

PR #101, opened 2020-08-20
First stab at retries and DLQ notifications for the shipper

Mpdreamz:
This is the first stab at actually handling the bulk responses in `Elastic.Ingest` and implement (configurable) backoffs. 

This also extends the callbacks available to be notified of various failures allowing you to create fallback/DLQ functionality on top of `ElasticsearchChannel<T>`. 

Pretty chuffed how well https://www.nuget.org/packages/Elasticsearch.Net.VirtualizedCluster/ worked to unit test all the moving pieces:
https://github.com/elastic/ecs-dotnet/pull/101/files#diff-840c42275ec505873ebfc4211bf72cd7R8

We are currently on a preview build of `Elasticsearch.Net.VirtualizedCluster` as we make use of some new features not yet released.


tests/Elastic.Ingest.Tests/Setup.cs
```diff
@@ -0,0 +1,102 @@
...
+
+			public BufferOptions<Base> BufferOptions { get; }
+
+			public ManualResetEventSlim WaitHandle { get; } = new ManualResetEventSlim();
+
+			public int Rejections => _rejections;
+			public int TotalBulkRequests => _requests;
+			public int TotalBulkResponses => _responses;
+			public int TotalRetries => _retries;
+			public int RetryDepletions => _retryDepletions;
```
russcam, line 76: Maybe `MaxRetriesExceeded`? I had to look up when `RetryRejectionCallback` is called 😃 Following this, perhaps 

`RetryRejectionCallback` -> `MaxRetriesExceededCallback` 

too?

Mpdreamz: Much better!

Mpdreamz:
> I noticed also that BufferOptions<TEvent> has references to LogEvent type in XML doc comments, which should now be TEvent

Thanks for the review @russcam , also updated the mention of `LogEvent` in the xml docs.

PR #108, opened 2020-10-22
fix #87 hostname shouldn't be written in server.IP field

### LINKED ISSUE #87: [BUG] Serilog event using host name instead of server IP
**ECS schema version (e.g. 1.4.0)**: 1.5.0

**ECS .NET assembly version (e.g. 1.4.2)**: 1.5.1 (Elastic.CommonSchema.Serilog)

**Elasticsearch version (if applicable)**: 7.4.0

**.NET framework / OS**: .NET Core 3.1 / macOS

When a request event contains a host name, it is used to override the server IP. The field server.ip is then invalid according to the ECS schema and the event cannot be indexed into Elasticsearch ("failed to parse field [server.ip] of type [ip]"). Since the field contained a valid IP before being overridden (at least with .NET Core) the expected behavior would be not to override it.

There might be some use cases I am not seeing here but at least using ASP.NET Core this cannot seem to work. The IP is overridden here: https://github.com/elastic/ecs-dotnet/blob/master/src/Elastic.CommonSchema.Serilog/LogEventConverter.cs#L195 with a value that comes from here: https://github.com/dotnet/aspnetcore/blob/5b2f3fb5f7f24ac3e91c5150a55cc411b2b36b76/src/Hosting/Hosting/src/Internal/HostingRequestStartingLog.cs#L33. For .NET Core using a HttpContextAcessor this IP is already set here and the value seems correct: https://github.com/elastic/ecs-dotnet/blob/master/src/Elastic.CommonSchema.Serilog/Http/HttpContextAccessorAdapter.cs#L133. So this might be a fix made for another framework. I don't have visibility on that.

**Steps to reproduce**:
 1. It works on a simple ASP.NET project so do a `dotnet new web` then add Elastic.CommonSchema.Serilog
 2. Setup a `EcsTextFormatterConfiguration` with `MapHttpContext`
 3. `dotnet run`
4. `curl localhost:5000` => The log line from `Microsoft.AspNetCore.Hosting.Diagnostics` should contain a `server.ip = localhost:5000` instead of an IP address

serdarkalayci:
Writing hostname to the IP field creates logs like the one below:

{"@timestamp":"2020-10-21T14:02:51.7698395+03:00","log.level":"Information","message":"Request starting HTTP/1.1 GET http://10.132.4.163:5111/readiness ","_metadata":{"message_template":"{HostingRequestStartingLog:l}","protocol":"HTTP/1.1","method":"GET","content_type":null,"content_length":null,"scheme":"http","path_base":"","path":"/readiness","query_string":"","hosting_request_starting_log":"Request starting HTTP/1.1 GET http://10.132.4.163:5111/readiness ","event_id":{"id":1},"request_id":"0HM3LHKKTLRML:00000001","request_path":"/readiness","span_id":"|3d1c3928-4b1f17b80024b28b.","trace_id":"3d1c3928-4b1f17b80024b28b","parent_id":"","connection_id":"0HM3LHKKTLRML"},"ecs":{"version":"1.5.0"},"event":{"severity":2,"timezone":"GMT+03:00","created":"2020-10-21T14:02:51.7698395+03:00"},"log":{"logger":"Microsoft.AspNetCore.Hosting.Diagnostics","original":null},"process":{"thread":{"id":25},"pid":1,"name":"dotnet","executable":"dotnet"},"server":{"address":"10.132.4.163:5111",**"ip":"10.132.4.163:5111"**}}

which is not parsable as IP
"failed to parse field [server.ip] of type [ip] in document with id '3TwkS3UBbmzuC991IrD5'. Preview of field's value: '10.132.4.163:5111'","caused_by":{"type":"illegal_argument_exception","reason":"'10.132.4.163:5111' is not an IP string literal."}

### DIFF
--- a/src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
+++ b/src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
@@ -194,7 +194,6 @@ namespace Elastic.CommonSchema.Serilog
 
 			var hasHost = e.TryGetScalarPropertyValue(SpecialKeys.Host, out var host);
 			server.Address = hasHost ? host.Value.ToString() : null;
-			server.Ip = hasHost ? host.Value.ToString() : null;
 			return server;
 		}
 


Mpdreamz (approved):
LGTM, host can be an IP but we should treat this as incidental.

Mpdreamz:
@serdarkalayci mind signing the https://www.elastic.co/contributor-agreement so that I may pull these bits in?

Mpdreamz:
Thanks for signing @serdarkalayci (and the fix of course 😄)

PR #112, opened 2020-11-04
Downgrade Microsoft.Extensions.* packages to 2.1.0.

sgryphon:
As per the GA plan, use lower level package for wider usability.


Mpdreamz (approved):
LGTM!

PR #107, opened 2020-10-16
Feature/standard options pattern 2

sgryphon:
Cleans up some of the Options pattern in the Extensions Logging to more follow the way things like OptionsMonitor<>, etc work.

Options pattern is meant for configuration settings, mostly passed in from external sources, e.g. appsettigns.json, enviornment variables, command line, etc, so is limited to values only.

Configuration for things like callbacks, which can be done in code, are in a separate function.

It includes an example that add the logger provider with a single line of code (plus a using statement), to show simple usage; and keeps the more advanced using (with callbacks, etc) as a separate example.

PR #114, opened 2020-11-16
Bug fix - logger not including custom template & scope

sgryphon:
I noticed that the Microsoft.Extensions logger was no longer including the custom properties for MessageTemplate and Scopes.

Changes in Base meant I needed to implement the template method to explicitly add the additional properties.

This change is really only the last commit, and can be cherry picked to another branch if needed.

It was done on the Options Pattern branch because that is where I have a working example/test project. The other example project has hard coded to ipv4.fiddler:9200, which doesn't exist on my machine. On this branch tests default to the default install location localhost:5601, and can be overridden via properties, e.g. from command line.

This should be merged before releasing Extensions GA #95


Mpdreamz:
@sgryphon mind rebasing this now that #107 made it in the main integration branch.

sgryphon:
Good to go now

Mpdreamz (commented):
Thank you for catching this! 

Is it worth implementing `TryRead` and `ReceiveProperty` to so that `LogEvent` property roundtrips during (de)serialization?

russcam:
I added an additional test to assert that the `MessageTemplate` and `Scopes` properties are deserialized by the Elasticsearch client; they both need `[DataMember]` attributes for the client to deserialize correctly, as the client's internal serializer looks for camel case field names in JSON by default, and is strict about casing.

russcam:
@sgryphon kicked off Jenkins to test in CI

russcam:
@sgryphon CI is green- is this ready to be merged?

sgryphon:
Yes, definitely. This one was actually a bug, so good to get it fixed.

Thanks for adding the help and adding the test (I think you added it).

Sly

PR #117, opened 2020-11-19
Fixed Exception StackTrace formatting to handle missing StackFrames

### LINKED ISSUE #102: [BUG]CatchErrors in LogConverter - stackframe can be null and is not treated in code
In CatchErrors method of LogConverter for Serilog you are trying to get first frame from the stacktrace of each exception (and also for all inner exceptions).
The problem is that the frame is not tested for null and it will throw an exception when trying to obtain the fileName, method etc.

To reproduce is very simple. In user code (in a controller for instance) just add this line:
throw new ApplicationException("lorem...", new ApplicationException("ipsum"));

Obs.: For the inner exception the stack trace will have no frames as it was issued in the same context like the outer exception and this will cause the LogConverter to fail on CatchErrors methon when treating the inner exceptions

snakefoot:
Resolves #102

### DIFF
--- a/src/Elastic.CommonSchema.NLog/README.md
+++ b/src/Elastic.CommonSchema.NLog/README.md
@@ -32,7 +32,11 @@ filesystem target and [Elastic Filebeat](https://www.elastic.co/downloads/beats/
   </extensions>
   <targets>
     <target name="console" type="console">
-      <layout xsi:type="EcsLayout" />
+      <layout xsi:type="EcsLayout">
+        <metadata name="MyProperty" layout="MyPropertyValue" /> <!-- repeated, optional -->
+        <label name="MyLabel" layout="MyLabelValue" />          <!-- repeated, optional -->
+        <tag layout="MyTagValue" />                             <!-- repeated, optional -->
+      </layout>
     </target>
   </targets>
   <rules>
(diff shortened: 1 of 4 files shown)


src/Elastic.CommonSchema.NLog/README.md
```diff
@@ -41,6 +45,54 @@ filesystem target and [Elastic Filebeat](https://www.elastic.co/downloads/beats/
 </nlog>
 ```
 
+## EcsLayout Parameter Options
```
Mpdreamz, line 48: Thank you for adding this!

Mpdreamz (approved):
LGTM 👍

src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
```diff
@@ -350,9 +352,11 @@ private static string CatchErrors(IReadOnlyCollection<Exception> errors)
 					fullText.WriteLine($"\tSource: {exception.TargetSite?.DeclaringType?.AssemblyQualifiedName}");
 					fullText.WriteLine($"\tMessage: {exception.Message}");
 					fullText.WriteLine($"\tTrace: {exception.StackTrace}");
-					fullText.WriteLine($"\tLocation: {frame.GetFileName()}");
-					fullText.WriteLine(
-						$"\tMethod: {frame.GetMethod()} ({frame.GetFileLineNumber()}, {frame.GetFileColumnNumber()})");
```
vrabac74, line 355: I got a null pointer exception when frame is used. Should it be checked for null ?

PR #136, opened 2021-03-22
Add version attribute entry

russcam:
This commit adds a version attribute entry to the docs,
and substitutes the value into code blocks.


bmorelli25 (approved):
Looks great. Thank you!

PR #110, opened 2020-10-29
Further Elastic.Ingest Improvements

Mpdreamz:
This builds upon #109 where `Elastic.Ingest` moves to the product neutral dependency `Elastic.Transport`. 

This PR introduces the following:

* `Elastic.Ingest` is now the base library for anything that wants to ship buffered data partioned by buffer size and max buffer age.
* `Elastic.Ingest.Transport` is now the base for anything that wants to push data using `Elastic.Transport`
	  * `Elastic.Ingest.Elasticsearch` allows any object to pushed to any index.
	  	 This PR introduces 2 Elasticsearch channels `IndexChannel` and a `DataStreamChannel` to provide the best experience for both without trying to abstract a single channel to work for both.
	  * `Elastic.Ingest.Apm` only allows apm events to be pushed to apm server.
* `Elastic.Ingest.OpenTelemetry` is a toy library that allows you to write `Activity`'s to an OpenTelemetry receiver using `TraceChannel` 

This PR also splits the internal channel into an inbound and outbound channel. The inbound channel's sole purpose it to batch based on size/lifetime. It then sends pages of data to the outbound channel where data is sequentially received and send out concurrently. 

Example usage of `ApmChannel` pushing transactions:

```csharp

var channelOptions = new ApmChannelOptions
{
	BufferOptions = options,
	ShipTo = new ShipTo(transport)
};

var channel = new ApmChannel(channelOptions);

string Id() => RandomGenerator.GenerateRandomBytesAsString(8);
var random = new Random();

for (var i = 0; i < numberOfEvents; i++)
{
	channel.TryWrite(new Transaction("http", Id(), Id(), new SpanCount(), random.NextDouble() * random.Next(100, 1000) , Epoch.UtcNow) { Name = "x" });
}
```

image

Opening early for discussing the approach, plenty of refactoring and actual implementations need to follow.

PR #160, opened 2021-07-27
Revert shipping Elastic.Ingest

Mpdreamz:
Still waiting on some improvements from #110 to come through before this should go live on NuGet.


Mpdreamz:
cc @elastic/dotnet self reviewed this one.

russcam:
Elastic.Ingest is a dependency for [Elasticsearch.Extensions.Logging](https://www.nuget.org/packages/Elasticsearch.Extensions.Logging/1.6.0-alpha1), which I think we would want to release the 1.6.0 GA version of soon.

PR #182, opened 2022-05-26
fix: deserialize nested json

santhoshar-bt:
BaseJsonConverter was throwing an error "Error: Exception occurred while processing event The converter 'Elastic.CommonSchema.Serialization.BaseJsonConverter' read too much or not enough." when trying to deserialize a nested json.

So adding in code fix to handle reading nested json.


Mpdreamz:
Hi @santhoshar-bt thank you for the PR can you describe what `nested json` means here?

If you could include a unit test with an example of json that was breaking the serialization that would be great!

Mpdreamz:
Caught this in a reproducible unit test in #254 and included a fix over there so closing this PR. 

:pray: Thank you for alerting us to this @santhoshar-bt and submitting the fix!

PR #190, opened 2022-07-21
Fix: double underscore in case uppercase character is preceeded by an underscore

andreycha:
**Problem**
Currently property like `User_Id` is serialized as `user__id` (note double underscore).

**Changes**
- Fix to produce only one underscore
- Add tests for snake case


Mpdreamz (approved):
LGTM thanks for catching this @andreycha ! Waiting to sort out the CLA before merging. 

Can you retarget this PR to the `main` branch?

andreycha:
Done.

Probably it's worth marking somewhere that once https://github.com/dotnet/runtime/issues/782 is released (seems to be part of .NET 7) custom naming policy can be dropped.

PR #192, opened 2022-07-25
Skip RequestPath property

andreycha:
Request logging middleware from Serilog.AspNetCore [adds `RequestPath` property](https://github.com/serilog/serilog-aspnetcore/blob/c15ab6d7a4e3100ae54f92352a4f69910cda8730/src/Serilog.AspNetCore/AspNetCore/RequestLoggingMiddleware.cs#L96) to log events.

This PR skips `RequestPath` property, it's already mapped to `url.path` if `MapHttpAdapter` is enabled (which is a preferred way of getting all the HTTP-related information in the log entry): https://github.com/elastic/ecs-dotnet/blob/main/src/Elastic.CommonSchema.Serilog/Http/HttpContextAccessorAdapter.cs#L108


andreycha:
As I dug a little bit deeper into the topic, this PR seems questionable to me (or at least incomplete). Both Serilog.AspNetCore and standard AspNet.Core loggers enrich events with request-related information which currently ends up in `metadata`. By defaults it's not propagated to standard ECS fields. Once user configures `MapHttpAdapter`, this information appears in ECS fields, but then it's becomes redundant in `metadata`.

Here is the summary (probably still incomplete):

Serilog.AspNetCore provides:
| Property  | ECS field | Is mapped |
| ------------- | ------------- | ------------- |
| RequestMethod | http.request.method | When MapHttpAdapter is configured |
| RequestPath | url.path | When MapHttpAdapter is configured |
| StatusCode | http.response.status_code | When MapHttpAdapter is configured |
| Elapsed | event.duration? | No |

Microsoft loggers provide:
| Property  | ECS field | Is mapped |
| ------------- | ------------- | ------------- |
| Protocol | http.version | No |
| Method | http.request.method | When MapHttpAdapter is configured |
| Scheme | url.scheme | When MapHttpAdapter is configured |
| PathBase | N/A | N/A |
| Path | url.path | When MapHttpAdapter is configured |
| QueryString | url.query | When MapHttpAdapter is configured |
| RequestId | http.request.id | No |

I guess there are multiple options of how `Elastic.CommonSchema.Serilog` might behave:
1) If `MapHttpAdapter` is not configured, populate `http` and `url` fields from request-related log properties.
2) If `MapHttpAdapter` is configured, let it populate `http` and `url` fields.
3) Merge together information available in log properties and provided by http adapter (if configured). This might be the best option because it will also cover fields which are currently not mapped.

In all cases it probably then makes sense to not duplicate request-related properties in `metadata` as it's already the case with other properties and enrichers.

The part with Microsoft loggers seems to be applicable to `Elastic.CommonSchema.NLog` as well.

Mpdreamz (approved):
LGTM! 

Waiting on the CLA before merging this in. Can you retarget this PR to `main` ?

andreycha:
Done.

Mpdreamz:
@andreycha that would be great! I wonder why we don't assign our `HttpAdapter` always? 

`Elastic.CommonSchema.NLog` and `Elasticsearch.Extensions.Logging` will need similar treatment but lets address those in a follow up PR.

Mpdreamz:
`Elastic.CommonSchema.NLog` is addressed by https://github.com/elastic/ecs-dotnet/pull/195/files

andreycha:
Alright, I will then prepare a separate PR.

> I wonder why we don't assign our `HttpAdapter` always?

It would be nice to have it always mapped but when should the library take `IHttpContextAccessor` from? And what if the host application is not a web application?

andreycha:
Closing this since https://github.com/elastic/ecs-dotnet/pull/213 has been merged.

PR #194, opened 2022-07-28
Update ECS.NET to latest ECS specification

Mpdreamz:
This PR updates the code generator to be able to handle the latest ECS specification format. 

ECS 1.6 introduced some changes in how fields and more importantly fields reuse were surfaced. 

While we had a beta for 1.6 around for a long time thanks to the painstaking work from @russcam on #145 the patches made were not enough to withstand the more recursive fields reusages introduced in 1.7 and above. 

This PR bites the bullet on @russcam's comment on the PR:

> The generation approach could benefit from separating a model used to generate the C# types from the ECS schema deserialized types, allowing an intermediate transformation process to transfrom from the deserialized types to a model that simplifies code generation.

And now introduces that intermediary canonical model from which we can generate the new code. 

A whole new [`Elastic.CommonSchema.Generator`](
https://github.com/elastic/ecs-dotnet/tree/fix/update-to-latest-spec/src/Elastic.CommonSchema.Generator) is now introduced replacing the older `Generator`. Although it reuses bits from the old one it now goes through an additional phase that projects the ECS schema [into a new model](https://github.com/elastic/ecs-dotnet/tree/fix/update-to-latest-spec/src/Elastic.CommonSchema.Generator/Domain). 

The new code generation brings a few breaking changes. 

Firstly `Base` entity is renamed to `EcsDocument` in the output to types. 

Each ECS fieldset is projected as partial class. 
Each field set is closed over as an entity that extends the fieldset and included references to other fieldsets as additonal properties. 

E.g the [`cloud`](https://www.elastic.co/guide/en/ecs/current/ecs-cloud.html) fieldset is translated into:

```csharp
abstract class CloudFieldSet {}
class Cloud : CloudFieldSet {
	public CloudOrigin Origin { get; }
	public CloudTarget Target { get; }
}
class CloudOrigin : CloudFieldSet {}
class CloudTarget : CloudFieldSet {}
```

This ensures fields can not be arbitrary nested e.g as `cloud.origin.origin.origin.origin`. 

Another big improvement is that we are much more conservative when to introduce `InlineObjects`. 

For example the [`log`](https://www.elastic.co/guide/en/ecs/current/ecs-log.html) field set defines the following fields:

- [log.origin.file.line](https://www.elastic.co/guide/en/ecs/current/ecs-log.html#field-log-origin-file-line)
- [log.origin.file.name](https://www.elastic.co/guide/en/ecs/current/ecs-log.html#field-log-origin-file-name)

Previously we would generate several inline objects to describe these: `LogOrigin` and `File`. In this update these exists directly as properties on `Log` since the ECS spec does not explicitly define `log.origin` or `log.origin.file` as objects. 

This PR further removes `Elastic.CommonSchemaNEST` since updating it was quite intensive and we have a brand spanking new client to target soon instead thanks to @stevejgordon!

Opening the PR since its fully functional and all tests pass again. Will keep amending  refactoring/renames as I find them and do a final pass.


src/Elastic.CommonSchema.Generator/Schema/DTO/Field.cs
```diff
@@ -162,5 +151,18 @@ public class Field
...
+		///  If <see cref="Type"/> is set to <see cref="FieldType.ScaledFloat"/> this configures the scaling factor
+		/// </summary>
+		[JsonProperty("scaling_factor")]
+		public int? ScalingFactor { get; set; }
+
+		/// <summary>
+		///  Validation (regex) pattern
+		/// </summary>
+		[JsonProperty("pattern")]
+		public string Pattern { get; set; }
```
gregkalapos, line 165: It seems there is a typo in the spec.

`client.mac` has "patther" instead of "pattern" [here](https://github.com/elastic/ecs/blob/main/generated/ecs/ecs_nested.yml#L484). From what I understand currently has 2 effects:
- We loose the defined pattern for the `client.mac` field
- The generator prints a warning

But the generator still works, so it's good.

Nevertheless, I think we should do something here.

Options I see:
- we open a PR to fix it, and the current version is just what it is and we accept it (I think this would be the best)
- we try to workaround and also parse "patthern" (I'd rather not...).

Mpdreamz: ++ already addressed this separately: 

https://github.com/elastic/ecs/pull/2014

Having the specification validation really paid off!

eugene-arg:
any plans to publish npm packages with the latest changes?

startaoneek:
@Mpdreamz, @gregkalapos  hi, any plans to publish new version of nuget packages with these changes?

Mpdreamz:
We do we just created https://github.com/elastic/ecs-dotnet/issues/197 to track our path towards getting this out there. You can subscribe to that ticket for updates. We can't commit to a specific date though.

src/Elastic.CommonSchema/EcsDocument.Generated.cs
```diff
@@ -0,0 +1,223 @@
...
+namespace Elastic.CommonSchema
+{
+	///<inheritdoc cref="BaseFieldSet"/>
+	public partial class EcsDocument : BaseFieldSet 
+	{
+
+		/// <summary>
+		/// Elastic Common Schema version v8.3.1
+		/// </summary>
+		public static string Version => "v8.3.1";
```
andreycha, line 30: Was that intended that before version had format "1.2.3" and now "v1.2.3"? Although [specification](https://www.elastic.co/guide/en/ecs/8.3/ecs-ecs.html) doesn't explicitly mention required format, example there uses "1.2.3" format.

PR #201, opened 2022-08-28
Add log4net support

andreycha:
This PR adds support for log4net via ECS layout.

Changes:
* Added log4net ECS layout
* Added unit tests
* Updated site documentation

Please let me know if I missed anything.

Some open questions:
- [x] Naming of the library: `Elastic.CommonSchema.Log4net` or `Elastic.CommonSchema.log4net`? Logging library is officially spelled as `log4net`. => `Elastic.CommonSchema.Log4net`
- [x] Do we want to provide the possibility to enrich logging event before logging? Similar to [NLog layout implementations](https://github.com/elastic/ecs-dotnet/blob/main/src/Elastic.CommonSchema.NLog/EcsLayout.cs#L154). Though this is only useful in case of programmatical configuration which I believe is a small minority of the use cases. => Not yet
- [x] Currently layout populates `service.version` field, but a wrong value might end up there. E.g., in case of IIS hosted applications `GetEntryAssembly()` returns null so version of the calling assembly is used which might be different from the application version. Keep or drop? => Drop
- [x] Is example application needed? => No


src/Elastic.CommonSchema.Log4net/LoggingEventConverter.cs
```diff
@@ -0,0 +1,197 @@
...
+			Event = GetEvent(loggingEvent),
+			Error = GetError(loggingEvent),
+			Service = GetService(),
+			Process = GetProcess(loggingEvent),
+			Host = GetHost(loggingEvent),
+			Metadata = GetMetadata(loggingEvent)
+		};
+
+	private static Log GetLog(LoggingEvent loggingEvent)
+    {
```
Mpdreamz, line 35: nit: use tabs instead of spaces

src/Elastic.CommonSchema.Log4net/EcsLayout.cs
```diff
@@ -0,0 +1,25 @@
...
+/// Formats log events into a JSON representation that adheres to Elastic Common Schema specification
+/// </summary>
+public class EcsLayout : LayoutSkeleton
+{
+    public override string ContentType => "application/json";
+
+	public override void ActivateOptions() => IgnoresException = false;
+
+	public override void Format(TextWriter writer, LoggingEvent loggingEvent)
+    {
```
Mpdreamz, line 21: nitpick: use tabs instead of spaces (reformat should do so).

Mpdreamz (approved):
LGTM! Left a few nitpicks to address around formatting.

Mpdreamz:
Jenkins run tests please

Mpdreamz:
- Naming of the library: Elastic.CommonSchema.Log4net or Elastic.CommonSchema.log4net? Logging library is officially spelled as log4net.

I have no strong opinion but it seems common for the suffix to be `Log4Net` on nuget.org. I think my vote would go to adhere to regular namespacing rules.

- Do we want to provide the possibility to enrich logging event before logging? Similar to [NLog layout implementations](https://github.com/elastic/ecs-dotnet/blob/main/src/Elastic.CommonSchema.NLog/EcsLayout.cs#L154). Though this is only useful in case of programmatical configuration which I believe is a small minority of the use cases.

I'm fine leaving this out until we see strong evidence this is needed.

- Currently layout populates service.version field, but a wrong value might end up there. E.g., in case of IIS hosted applications GetEntryAssembly() returns null so version of the calling assembly is used which might be different from the application version. Keep or drop?

Drop, best to allign with what the Elastic Apm Agent does here:

https://github.com/elastic/apm-agent-dotnet/blob/305be0a9017ce9be50b4b269caa7404d14974cbb/src/Elastic.Apm/Config/AbstractConfigurationReader.cs#L853-L860

- Is example application needed?

No, the readme covers everything already IMO.

andreycha:
> Drop, best to allign with what the Elastic Apm Agent does here:
> https://github.com/elastic/apm-agent-dotnet/blob/305be0a9017ce9be50b4b269caa7404d14974cbb/src/Elastic.Apm/Config/AbstractConfigurationReader.cs#L853-L860

Dropped for now. Imho best would be to implement later this approach for all three packages.

src/Elastic.CommonSchema.Log4net/EcsLayout.cs
andreycha: Thanks for noticing. Done.

src/Elastic.CommonSchema.Log4net/LoggingEventConverter.cs
andreycha: Done.

Mpdreamz:
```
21:13:55  [xUnit.net 00:00:01.00]     Elastic.CommonSchema.Log4net.Tests.MessageTests.ToEcs_AnyEvent_PopulatesHostField [FAIL]
21:13:55    Failed Elastic.CommonSchema.Log4net.Tests.MessageTests.ToEcs_AnyEvent_PopulatesHostField [36 ms]
21:13:55    Error Message:
21:13:55     Expected info.Host.Hostname to be 
21:13:55  "APM-CI-IMMUTABL" with a length of 15, but 
21:13:55  "apm-ci-immutable-windows-2019-1662404522828227605" has a length of 49, differs near "apm" (index 0).
```

It seems log4net lowercases hostname?

andreycha:
> It seems log4net lowercases hostname?

Not really. It determines hostname using several sources in certain order. Whereas in the test I just compared the value against `Environment.MachineName` which returns hostname in upper case and truncated to 15 characters (I wasn't aware of such behavior). Fixed the test.

Mpdreamz:
Merged, thanks for donating this implementation @andreycha :+1:

PR #204, opened 2022-08-30
Serilog: refactor url and server population

### LINKED ISSUE #198: Query string is not included in the logs
**ECS version**: 1.5.0

**Elastic.CommonSchema.Serilog version**: 1.5.3

**.NET framework / OS**: ASP.NET Core 6

The query string is not logged in the url field of the log.

In Elastic.CommonSchema.Serilog.HttpAdapter.ConvertToUri() an instance of an Uri is created from the HttpRequest:

new Uri($"{request.Scheme}://{request.Host}{request.Path}");

The query string in the request is not included so Uri.Query will be null. 

Therefore the url returned from the propery Url will not have the [Query property](https://github.com/elastic/ecs-dotnet/blob/main/src/Elastic.CommonSchema.Serilog/Http/HttpContextAccessorAdapter.cs#L111) set.

**Steps to reproduce**:
 1. Call an endpoint with query parameters.

andreycha:
- Fix query string population according to the [specification](https://www.elastic.co/guide/en/ecs/1.12/ecs-url.html#field-url-query) (closes #198)
- Use only `HttpRequest` to populate `url` and `server` fields, without creation of `Uri` instance
- Put original full URL into `url.original` instead of just path (according to the [spec](https://www.elastic.co/guide/en/ecs/1.12/ecs-url.html#field-url-original))
- Put host name into `url.domain` and `server.domain`


Mpdreamz (approved):
LGTM

PR #206, opened 2022-08-31
Serilog Sink based on CommonSchemaChannel

Mpdreamz:
- stage
- New Serilog Sink


Mpdreamz:
Merging this in because its been open for a long time and is in a decent enough shape for the alpha (with unit/integration tests). 

Needs documentation though.

Issue #161, opened 2021-08-02
[FEATURE] Turn off/Override Snake Case Serialization of Metadata


ghost:
Metadata properties and classes at the top level get renamed with underscores.

For instance:

If I add a class called "ContextInfo" to the LogEventInfo properties collection.  It doesn't matter what the property name is in the dictionary.  The resultant JSON is this has "context_info" instead.
"metadata" :{
  "context_info" : {
 "original_id" : 123345
}
}

Class properties can be controlled with serialization attributes: [JsonProperty("originalId")].  With that I can fix "original_id" to display as "originalId" in the JSON.  

This is causing frustrations with the development team and preventing better adoption of this format.

Can we get better control over the serialization naming?

Mpdreamz:
I agree looking back snake casing the metadata keys violates the principle of least surprise

I've opened: https://github.com/elastic/ecs-dotnet/pull/211 to address this. 

It will be a breaking change for 8.4.0. 

Thanks for opening the issue

PR #211, opened 2022-09-01
Fix Metadata serialization

Mpdreamz:
Metadata now has a dedicated Dictionary implementation that reads and writes objects verbatim.

Previously we registered a global Dictionary json converter for every Dictionary, we now only use a custom serialization routine for Metadata.

This ensures metadata is sent and received verbatim and no longer forcefully snake_cases the keys. This introduced too much magic and violated to principle of least surprise.

Fixes https://github.com/elastic/ecs-dotnet/issues/161

Breaking change for 8.4.0 but one I think is worth shouldering.


src/Elastic.CommonSchema/Serialization/MetadataDictionaryConverter.cs
```diff
@@ -0,0 +1,89 @@
...
+			{
+				var propertyName = kvp.Key;
+				writer.WritePropertyName(propertyName);
+
+				if (kvp.Value == null)
+					writer.WriteNullValue();
+				else
+				{
+					var inputType = kvp.Value.GetType();
+					//TODO prevent reentry and cache get converters
```
z1c0, line 56: Is this intended for a future PR?

z1c0 (approved):
LGTM!

(I meant to ask about compatibility but I see you answered that already in the PR description)

src/Elastic.CommonSchema/Serialization/MetadataDictionaryConverter.cs
Mpdreamz: No that's a stale TODO in older versions of System.Text.Json it was worthwhile performance wise to call directly into any known converter:

https://github.com/elastic/ecs-dotnet/pull/211/files/229ac063e5981c3305a6bb72f11a49e1be73ca94#diff-ed861ef4bf6340af5d35ed6493d67a8f071fdd6cb2c80d125cbc4333499d3f5eL73-L74

However this no longer works nor has any benefit over calling `.Deserialize<>()`.

PR #212, opened 2022-09-01
user_agent.name should contain browser name

### LINKED ISSUE #174: [BUG] Incorrect Value on UserAgent Name Field
**ECS integration/library project(s) (e.g. Elastic.CommonSchema.Serilog)**: Elastic.CommonSchema.Serilog

**ECS schema version (e.g. 1.4.0)**: All

**ECS .NET assembly version (e.g. 1.4.2)**: Since the initial version

**Elasticsearch version (if applicable)**: N/A

**.NET framework / OS**: All supported versions

`name` subfield of `user_agent` field should be filled only with the browser name, e.g. _Firefox_, according to ECS specification. However, in current implementation, this field is filled with full user agent string, e.g. _Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:94.0) Gecko/20100101 Firefox/94.0_.

**Steps to reproduce**:
 1. Map the HttpContextAccessor to the `EcsTextFormatterConfiguration`.
 2. Trigger a log.
 3. See that `user_agent.name` is filled with full user agent string.

image

I want to open a PR to correct this if it is ok for you.

andreycha:
Fix #174


Mpdreamz (approved):
LGTM Thank you for fixing this one!

src/Elastic.CommonSchema.Serilog/Http/HttpContextAccessorAdapter.cs
```diff
@@ -41,7 +41,7 @@ public UserAgent UserAgent
 
 				return new UserAgent
 				{
-					Name = userAgent,
+					Name = clientInfo.UA.Family,
```
Mpdreamz, line 44: `clientInfo.UA` never null?

andreycha: Honestly I was guided by usages of `clientInfo.UA` below where version is assigned. I checked now -- this approach appears to be correct as UAParser always fills those properties out (`UA`, `OS`, `Device`): https://github.com/ua-parser/uap-csharp/blob/eca6f57a2682be60984860e0675faabc6ef60d8d/UAParser/UAParser.cs#L294

PR #195, opened 2022-07-29
NLog EcsLayout with support for Http and Url properties

snakefoot:
With help from NLog.Web.AspNetCore (when available).

PR #213, opened 2022-09-02
Serilog: map more http and url information

andreycha:
**Description**

Following the discussion in https://github.com/elastic/ecs-dotnet/pull/192.

Both Serilog.AspNetCore and standard AspNet.Core loggers enrich events with http-related information which currently ends up in `metadata`. By defaults it's not propagated to standard ECS fields. Once user configures `MapHttpAdapter` some of this information appears in ECS fields (and becomes redundant in `metadata`).

This PR puts such values out of metadata to corresponding `http` and `url` fields.

**Summary of the properties**

Serilog.AspNetCore enriches events with:
| Property  | Target ECS field | Was mapped before | Is now mapped |
| ------------- | ------------- | ------------- | ------------- |
| RequestMethod | http.request.method | When MapHttpAdapter is configured | Yes |
| RequestPath | url.path | When MapHttpAdapter is configured | Yes |
| StatusCode | http.response.status_code | When MapHttpAdapter is configured | Yes |
| Elapsed | event.duration (with conversion to ns) | No | Yes |

Microsoft loggers enrich events with:
| Property  | Target ECS field | Was mapped before | Is now mapped now |
| ------------- | ------------- | ------------- |------------- |
| Protocol | N/A | No | No. Value can't be assigned directly. E.g. sample value is "HTTP/2" while `http.version` supposes values like `1.1` |
| Method | http.request.method | When MapHttpAdapter is configured | Yes |
| Scheme | url.scheme | When MapHttpAdapter is configured | Yes |
| PathBase | N/A | No | No |
| Path | url.path | When MapHttpAdapter is configured | Yes |
| QueryString | url.query | When MapHttpAdapter is configured | Yes |
| RequestId | http.request.id | No | Yes |
| ElapsedMilliseconds | event.duration (with conversion to ns) | No | Yes |
| ContentType | http.response.mime_type | No | No. Not clear whether it belongs to request or response. |
| StatusCode | http.response.status_code | When MapHttpAdapter is configured | Yes |

**Open questions**

- [ ] How can we reliably understand that for instance `Path` property is related to http request/response and is not just a custom user property? Have been thinking about checking logger name.


Mpdreamz (approved):
LGTM

- How can we reliably understand that for instance Path property is related to http request/response and is not just a custom user property? Have been thinking about checking logger name.

Can we infer this was set because other properties are set? e.g `Path` is never set in isolation by the default enrichers correct?

Mpdreamz:
Jenkins run tests please

andreycha:
> Can we infer this was set because other properties are set? e.g Path is never set in isolation by the default enrichers correct?

Yes, these properties come together. Doing this check for every property seems to be quite cumbersome. (Any of these properties might conflict with custom user properties.)

These log entries always come from certain loggers: `Serilog.AspNetCore.RequestLoggingMiddleware` and `Microsoft.AspNetCore.Hosting.Diagnostics`. Names are unlike to change so I thought it might be easier to check for logger name (similar to checking for NLog.Web assembly presence in https://github.com/elastic/ecs-dotnet/pull/195).

Mpdreamz:
I've merged this in as is. If you have the time to create a follow up PR to harden it with loaded assembly name checks that would be great!

PR #217, opened 2022-09-06
Allow EcsTextFormatter to accept any custom EcsDocument implementation

### LINKED ISSUE #167: [FEATURE] Allow custom base type
**ECS integration/library project(s) (e.g. Elastic.CommonSchema.Serilog)**:
Elastic.CommonSchema.Serilog

**Is your feature request related to a problem? Please describe.**
We are extending the ECS with our own fields. Let's use the Microsoft example company Contoso. We are extending the ECS with a new group named `contoso`. The current Serilog Converter doesn't support any Extension to the `Base` class.

The Method `ConvertToEcs` is implemented with the `Base` class. There is no other way to use the code in there with an other Type.
```csharp
public static Base ConvertToEcs(LogEvent logEvent, IEcsTextFormatterConfiguration configuration)
{
	... 
	var ecsEvent = new Base();
	...
	return ecsEvent;
}
```

**Describe the solution you'd like**
Without breaking changes we could at least change this to a method where it is possible to extend the `Base` class and add the own fields with casting in a other base type.
```csharp
public static Base ConvertToEcs(LogEvent logEvent, IEcsTextFormatterConfiguration configuration)
	=> ConvertToEcs<Base>(logEvent, configuration);
public static Base ConvertToEcs<T>(LogEvent logEvent, IEcsTextFormatterConfiguration configuration)
	where T : Base, new()
{
	... 
	var ecsEvent = new T();
	...
	return ecsEvent;
}
```

With this change we can use a own base class and easily add new fields.
```csharp
public class ContosoBase : Base
{
	[DataMember(Name = "contoso")]
	public Contoso Contoso { get; set; }
}
public class Contoso
{
	[DataMember(Name = "company_name")]
	public string CompanyName { get; set; }
}
public class ContosoEcsTextFormatter : EcsTextFormatter
{
	public override void Format(LogEvent logEvent, TextWriter output)
	{
		var ecsEvent = LogEventConverter.ConvertToEcs<ContosoBase>(logEvent, _configuration);
		if (ecsEvent is ContosoBase contosoEcsEvent)
		{
			contosoEcsEvent.Contoso = new Contoso
			{
				CompanyName = "Contoso",
			};
		}
		output.WriteLine(ecsEvent.Serialize());
	}
}
```

**Describe alternatives you've considered**
A problem here is the property `MapCustom` in the configuration of the formatter. 
```csharp
public interface IEcsTextFormatterConfiguration
{
	Func<Base, LogEvent, Base> MapCustom { get; set; }
	...
}
```
If we accept breaking changes, we could even change this and return the custom base type in the converter.
```csharp
public static T ConvertToEcs<T>(LogEvent logEvent, IEcsTextFormatterConfiguration configuration)
	where T : Base, new()
{
	... 
}
```

**Additional context**

What do you think about this change? Is it worth it to create a pull request?

Mpdreamz:
This way users can implement and use their own EcsDocument subclasses as `ITextFormatter`.

This introduces some breaking changes:

- `EcsTextFormatterConfiguration` is no longer a fluent descriptor, just a simple POCO.
- The MapFromContext() helper method no longer exists.

Fixes #167

### DIFF
--- a/examples/aspnetcore-with-serilog/Program.cs
+++ b/examples/aspnetcore-with-serilog/Program.cs
@@ -26,11 +26,11 @@ namespace AspnetCoreExample
 
 					// Ensure HttpContextAccessor is accessible
 					var httpAccessor = ctx.Configuration.Get<HttpContextAccessor>();
-					
+
 					// Create a formatter configuration to se this accessor
 					var formatterConfig = new EcsTextFormatterConfiguration();
-					formatterConfig.MapHttpContext(httpAccessor);
-					
+					formatterConfig.MapHttpAdapter = new HttpAdapter(httpAccessor);
+
 					// Write events to the console using this configration
 					var formatter = new EcsTextFormatter(formatterConfig);
 
--- a/src/Elastic.CommonSchema.Serilog/EcsTextFormatter.cs
+++ b/src/Elastic.CommonSchema.Serilog/EcsTextFormatter.cs
@@ -12,19 +12,28 @@ namespace Elastic.CommonSchema.Serilog
 	/// <summary>
 	/// A serilog formatter that writes log events using the Elasticsearch Common Schema format
 	/// </summary>
-	public class EcsTextFormatter : ITextFormatter
+	public class EcsTextFormatter<TEcsDocument> : ITextFormatter
+		where TEcsDocument : EcsDocument, new()
 	{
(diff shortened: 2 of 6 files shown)


sgtobin:
When and in what version will this change be available?

PR #218, opened 2022-09-06
Ensure ECS version is the schema version

### LINKED ISSUE #216: [QUESTION] ecs.version format has changed
**ECS version**: 8.3.1+

After https://github.com/elastic/ecs-dotnet/pull/194/ ECS version format has changed its format from `"1.2.3"` to `"v1.2.3"`: see [current value](https://github.com/elastic/ecs-dotnet/blob/main/src/Elastic.CommonSchema/EcsDocument.Generated.cs#L32). Although [the spec](https://www.elastic.co/guide/en/ecs/8.4/ecs-ecs.html#field-ecs-version) doesn't give any details, it uses `"1.2.3"` format as an example. Was this change intended?

Mpdreamz:
And not the gitref that we used to generate the types from

Fixes #216

### DIFF
--- a/src/Elastic.CommonSchema.Generator/Projection/TypeProjector.cs
+++ b/src/Elastic.CommonSchema.Generator/Projection/TypeProjector.cs
@@ -98,14 +98,14 @@ namespace Elastic.CommonSchema.Generator.Projection
 
 			Projection = new CommonSchemaTypesProjection
 			{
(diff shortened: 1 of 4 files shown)


Mpdreamz:
@gregkalapos @z1c0 self reviewing this FYI.

PR #222, opened 2022-09-10
Elastic.CommonSchema.NLog - MetadataDictionary with safe values

snakefoot:
System.Text.Json is very fragile, and can only handle safe objects


Mpdreamz:
Jenkins run tests please

Mpdreamz (approved):
Thanks for this update to support NLOG message templates defaults of `ToString()` with better support for `@` destructuring properties!

Mpdreamz:
Jenkins run tests please

PR #236, opened 2023-01-25
use GitHub actions (linux support)

v1v:
### What

- Use GitHub actions for the CI
- Report test results
- Export GH action data to OTEL
- Use PreCommit GH actions

### Follow up

- Support windows
- Release automation

PR #238, opened 2023-01-26
action: release

v1v:
### What

Use GH actions for the releases when:
- merge commits in the main branch
- or tag creation starting with `<major>.<minor>.<patch>`

### Tasks

- [x] Publish to nuget
- [x] Publish the releases


.github/workflows/docker-run/action.yml
```diff
@@ -0,0 +1,22 @@
+---
+name: common build tasks
```
v1v, line 2: This composite action is the one to be used in https://github.com/elastic/ecs-dotnet/pull/236/checks too

.github/workflows/release.yml
```diff
@@ -0,0 +1,127 @@
...
+    - if: ${{ failure() }}
+      uses: elastic/apm-pipeline-library/.github/actions/slack-message@current
+      with:
+        url: ${{ secrets.VAULT_ADDR }}
+        roleId: ${{ secrets.VAULT_ROLE_ID }}
+        secretId: ${{ secrets.VAULT_SECRET_ID }}
+        channel: "#apm-agent-dotnet"
+        message: |
+          :large_yellow_circle: [${{ github.repository }}] Release *${{ github.ref_name }}* could not be published.
+          Build: (<${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|here>)
```
v1v, line 127:
```suggestion
          Build: (<${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|here>)

```

pazone (approved):

pazone (approved):

Mpdreamz:
| or tag creation starting with <major>.<minor>.<patch>

Just double checking that this means we can publish `8.4.0-alpha1` just the same as always?

v1v:
> just the same as always?

`'[0-9]+.[0-9]+.[0-9]+.*' ` is the pattern I used so, it should support `8.4.0-alpha1` and `8.4.0` or `8.4.0-foo-bar`


Previously it was a bit fine defined with `tag pattern: '\\d+\\.\\d+\\.\\d+(-(alpha|beta|rc)\\d*)?', comparator: 'REGEXP'`

If you prefer, I can use the previous defined pattern so, we ensure it provides the same experience

Mpdreamz (commented):
LGTM.

Mpdreamz:
The only comment I have is it could be possible to combine these flows e.g:

https://github.com/elastic/elastic-ingest-dotnet/blob/main/.github/workflows/ci.yml

That way it's a tad easier to keep everything in sync and only one thing runs whether on a tag or not.

v1v:
> That way it's a tad easier to keep everything in sync and only one thing runs whether on a tag or not.

Let me propose a reusable action, so both workflows can benefit from that implementation and avoid the duplication

v1v:
Wait a minutes, I misunderstood your question.

We are using a different approach, one workflow per event or so, such as releases, or PRs. This will help us to distinguish the access to secrets since they cannot be accessed by defaults on PRs.

By doing so, there might be duplications, but will also avoid the usage of `if` conditions.

If you prefer one workflow to rule everything, we can try, though, we internally discussed to sue the mentioned approach for the shake of simplicity even though it might produce duplicated code.

What do you prefer?

.github/workflows/release.yml
```diff
@@ -0,0 +1,97 @@
...
+  DOTNET_VERSION: '6.0.400'
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+       fetch-depth: 0
```
Mpdreamz, line 20: I believe we need `fetch-depth` of `1` to satisfy our `MinVer` tooling being able to determine the appropiate version number based on available git tags.

Mpdreamz: See e.g https://github.com/elastic/elastic-ingest-dotnet/blob/main/.github/workflows/ci.yml#L46-L52

Mpdreamz (changes requested):

Mpdreamz:
| If you prefer one workflow to rule everything, we can try, though, we internally discussed to sue the mentioned approach for the shake of simplicity even though it might produce duplicated code.

| What do you prefer?

Whatever has us fall in line with the rest of the repositories @elastic/ci-robots  help us automate! :cat: 

I personally prefer one workflow so that I have more confidence whatever runs on tags/main integration branches is what was ran on PR's.

.github/workflows/release.yml
v1v: IIUC, `fetch-depth: 0` retrieves everything 

```
    # Number of commits to fetch. 0 indicates all history for all branches and tags.
    # Default: 1
    fetch-depth: ''
```

as explained in https://github.com/actions/checkout

I guess there is no need to fetch anything if `0`

v1v:
> I personally prefer one workflow so that I have more confidence whatever runs on tags/main integration branches is what was ran on PR's.

https://github.com/elastic/ecs-dotnet/pull/238/commits/a484ada1d9456a3671a1f705fb12209e1424bcb6 was the change I made to have one workflow per release type, while it was originally one workflow for both releases `main` or tags releases. 

@pazone and I discussed about the benefits and came up with havin several workflows with branches filter (`on: push: branches: - ...` ), but with simple actions. Rather than havin a single workflow and a powerful action handling all branches cases (like in the PR).

I'd somehow keep it as is if no strong opinion to merge both workflows as we are using the approach in other places.

Mpdreamz:
| I'd somehow keep it as is if no strong opinion to merge both workflows as we are using the approach in other places.

:+1:  Consistency trumps this style discussion

.github/workflows/release.yml
Mpdreamz: Ahh perfect!

Mpdreamz (approved):
LGTM

PR #246, opened 2023-02-01
action: use push for the tags

v1v:
tags filter is not available for `create` but `push` instead.

https://github.com/actions/runner/issues/1007


Mpdreamz (approved):
LGTM

PR #267, opened 2023-02-14
Improves on Elasticsearch.Extensions.Logging and ASP.NET

Mpdreamz:
Projects known labels to http.* and url.* now.

Improves how Scopes.* are resolved.

### DIFF
--- a/examples/aspnetcore-with-extensions-logging/Program.cs
+++ b/examples/aspnetcore-with-extensions-logging/Program.cs
@@ -1,0 +1,42 @@
+// -- Start an Elasticsearch Instance --
+
+using Elastic.Clients.Elasticsearch;
+using Elastic.Elasticsearch.Ephemeral;
+using Elasticsearch.Extensions.Logging;
+
+using var cluster = new EphemeralCluster("8.4.0");
+var client = CreateClient(cluster);
+//check if an instance is already running before starting
+if (!(await client.InfoAsync()).IsValidResponse)
+	cluster.Start(TimeSpan.FromMinutes(1));
+else Console.WriteLine("Using already running Elasticsearch instance");
+var builder = WebApplication.CreateBuilder(args);
+
+// Add services to the container.
+
+builder.Services.AddControllers();
+builder.Services.AddEndpointsApiExplorer();
+
+builder.Host.ConfigureLogging((_, loggingBuilder) =>
+{
+	loggingBuilder.AddElasticsearch(client.Transport, log =>
+	{
+		log.Tags = new[] { "debug" };
+	}, channel =>
+	{
+		channel.ResponseCallback = (response, buffer) => Console.WriteLine($"Written  {buffer.Count} logs to Elasticsearch: {response.ApiCallDetails.HttpStatusCode}");
+	});
+});
+var app = builder.Build();
+
+app.UseAuthorization();
+app.MapControllers();
+app.Run();
+
+
+static ElasticsearchClient CreateClient(EphemeralCluster cluster)
+{
+	var settings = new ElasticsearchClientSettings(cluster.NodesUris().First())
+		.EnableDebugMode();
+	return new ElasticsearchClient(settings);
+}
--- a/src/Elasticsearch.Extensions.Logging/ElasticsearchLogger.cs
+++ b/src/Elasticsearch.Extensions.Logging/ElasticsearchLogger.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 using Elastic.CommonSchema;
 using Elastic.Channels;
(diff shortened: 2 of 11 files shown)


examples/aspnetcore-with-extensions-logging/Program.cs
```diff
@@ -0,0 +1,42 @@
...
+var builder = WebApplication.CreateBuilder(args);
+
+// Add services to the container.
+
+builder.Services.AddControllers();
+builder.Services.AddEndpointsApiExplorer();
+
+builder.Host.ConfigureLogging((_, loggingBuilder) =>
+{
+	loggingBuilder.AddElasticsearch(client.Transport, log =>
```
HubDevUser, line 22: How would this work for AddElasticCloud ?

Mpdreamz: Hi @HubDevUser, There is an `.AddElasticCloud()` extension method too.
Although not one that takes `Transport` is that the question?

HubDevUser: When i use .AddElasticeCloud(CloudId, ApiKey) it does not report anything to my elastic cloud instance. It does work when i do .AddElasticCloud(CloudId, ApiKey) in this https://github.com/elastic/ecs-dotnet/blob/main/examples/console-with-extensions-logging/Program.cs#L25 console example though.  .AddElasticSearch() does work on my local dev docker image. How can I make this example work on a cloud instance :'( ?

HubDevUser: managed to work it 

	loggingBuilder.AddElasticsearch();

	void configure(ElasticsearchLoggerOptions options)
	{
		options.ShipTo.ConnectionPoolType = ConnectionPoolType.Cloud;
		options.ShipTo.CloudId = "id";
		options.ShipTo.ApiKey = "key";
	}

	builder.Services.Configure((Action<ElasticsearchLoggerOptions>)configure);

HubDevUser: this works L

var builder = WebApplication.CreateBuilder(args);

// Add services to the container.

builder.Services.AddControllers();
builder.Services.AddEndpointsApiExplorer();

builder.Host.ConfigureLogging((_, loggingBuilder) =>
{
	loggingBuilder.AddElasticsearch();

	void configure(ElasticsearchLoggerOptions options)
	{
		options.ShipTo.ConnectionPoolType = ConnectionPoolType.Cloud;
		options.ShipTo.CloudId = "id";
		options.ShipTo.ApiKey = "key";
	}

	builder.Services.Configure((Action<ElasticsearchLoggerOptions>)configure);

});
var app = builder.Build();

app.UseAuthorization();
app.MapControllers();
app.Run();



this does not :

namespace WebApplication5
{
    public class Program
    {
        public static void Main(string[] args)
        {
            CreateHostBuilder(args).Build().Run();
        }

        public static IHostBuilder CreateHostBuilder(string[] args) =>
            Host.CreateDefaultBuilder(args)
                .ConfigureWebHostDefaults(webBuilder =>
                {
                    webBuilder.ConfigureLogging((builder) =>
                    {
                        builder.AddElasticsearch();

                        void configure(ElasticsearchLoggerOptions options)
                        {
                            options.ShipTo.ConnectionPoolType = Elastic.Ingest.ConnectionPoolType.Cloud;
                            options.ShipTo.CloudId = "id";
                            options.ShipTo.ApiKey = "key";
                        }

                        builder.Services.Configure((Action<ElasticsearchLoggerOptions>)configure);
                    })
                    .UseStartup<Startup>();
                });
    }
}

src/Elasticsearch.Extensions.Logging/Options/ElasticsearchLoggerOptions.cs
```diff
@@ -68,5 +69,11 @@ public class ElasticsearchLoggerOptions : IEcsDocumentCreationOptions
 		/// 'Production', etc).
 		/// </summary>
 		public string[]? Tags { get; set; }
+
+		/// <summary>
+		/// Allows the direct setting of a <see cref="HttpTransport{TConfiguration}"/> to be used to communicate with Elasticsearch.
+		/// <para>If set takes precedence over <see cref="ShipTo"/> </para>
+		/// </summary>
+		public HttpTransport? Transport { get; set; }
```
sgryphon, line 77: @Mpdreamz  please don't do this, as it breaks the Options pattern. All options need to be simple types, so that they can be set from configuration sections, i.e. JSON.

I know I haven't looked at this for a while, but configuration options really need to be kept configurable, i.e. be able to be set through the configuration system (files, commands line, environment variables, Azure web app deployments, KeyVault, etc).  This basically means strings (environment variables are strings), or simple types.

Note that the default extension method `loggingBuilder.AddElasticsearch()` automatically uses the configuration section "Logging:Elasticsearch", so you don't need any special C# code, and everything can be done via config.

In some dynamic situations that may not be possible, which is fine, but in those cases use a different extension method / constructor. And if at all possible, also try and express the same options via configuration (strings).

Maybe add a comment to the file about preserving the Options pattern and only use strings, or simple types that can be expressed as strings (arrays and dictionaries are also allowed; basically whatever fits in JSON).

src/Elasticsearch.Extensions.Logging/ElasticsearchLogger.cs
```diff
@@ -61,50 +62,103 @@ public class ElasticsearchLogger : ILogger
...
-						logEvent.MessageTemplate = kvp.Value.ToString();
-						continue;
-					}
-					logEvent.AssignField(kvp.Key, FormatValue(kvp.Value));
-				}
+				logEvent.MessageTemplate ??= kvp.Value.ToString();
+				return;
+			}
+			var value = FormatValue(kvp.Value);
+			if (!AssignKnownHttpKeys(logEvent, kvp.Key, value))
```
sgryphon, line 106: This is kind of nice, i.e. if the scope values match the names set by ASP.NET, then st the logEvent.Http schema.

sgryphon: @Mpdreamz  maybe update the readme documentation to mention that these are automatically parsed and set.

https://github.com/elastic/ecs-dotnet/blob/main/src/Elasticsearch.Extensions.Logging/README.md

PR #274, opened 2023-02-23
action: test the new test reporter

v1v:
Test https://github.com/elastic/apm-pipeline-library/pull/2106

### DIFF
--- a/.github/workflows/test-reporter.yml
+++ b/.github/workflows/test-reporter.yml
@@ -12,7 +12,7 @@ jobs:
   report:
     runs-on: ubuntu-latest
     steps:
-      - uses: elastic/apm-pipeline-library/.github/actions/test-report@current
+      - uses: elastic/apm-pipeline-library/.github/actions/test-report@feature/update-test-reporter-action
         with:
           artifact: test-results        # artifact name
           name: Test Summary            # Name of the check run which will be created


.github/workflows/test-reporter.yml
```diff
@@ -12,7 +12,7 @@ jobs:
   report:
     runs-on: ubuntu-latest
     steps:
-      - uses: elastic/apm-pipeline-library/.github/actions/test-report@current
+      - uses: elastic/apm-pipeline-library/.github/actions/test-report@feature/update-test-reporter-action
```
v1v, line 15: To be changed with this branch so we can verify it. works as expected then we can revert it. This workflow needs to exist in the default branch

Mpdreamz (approved):
LGTM but I added a commit to un-comment the list-suites and list-tests which I think still requires a change in apm-pipeline-library:

https://github.com/elastic/apm-pipeline-library/pull/2106#issuecomment-1443296279

Mpdreamz:
I think there is still a bug in upstream test reporting:

https://github.com/phoenix-actions/test-reporting/issues/29

v1v:
https://github.com/elastic/apm-pipeline-library/pull/2106/files is now using your forked version

So if things work well, this is now solved temporarily

PR #279, opened 2023-03-01
Release to github package repository as well

Mpdreamz:
Sadly GH package repos still does not support anonymous access so we can't replace feedz.io with it yet. Feedz.io has also served us amazingly well for a long time so not in any rush replacing it. 

This additional push does help advertise which nuget packages this repository builds e.g:

image

From https://github.com/elastic/elastic-ingest-dotnet which already pushes to GH package repository.

### DIFF
--- a/.github/workflows/release-main.yml
+++ b/.github/workflows/release-main.yml
@@ -39,6 +39,7 @@ jobs:
 
     - name: publish canary packages github package repository
       shell: bash
+      timeout-minutes: 10
       run: |
         until dotnet nuget push 'build/output/*.nupkg' -k ${{secrets.GITHUB_TOKEN}} --skip-duplicate --no-symbols; do echo "Retrying"; sleep 1; done;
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release-main.yml
```diff
@@ -37,6 +37,11 @@ jobs:
         dotnet-sdk-version: ${{ env.DOTNET_VERSION }}
         command: ./build.sh generateapichanges -s true
 
+    - name: publish canary packages github package repository
+      shell: bash
```
v1v, line 41: What about a timeout if taking longer than X minutes?

```suggestion
      shell: bash
      timeout-minutes: 10
```

OTOH, will this step need to run before the other ones? If it fails, should fail always?

v1v (approved):
LGTM, just a minor comment

Mpdreamz:
@v1v thanks for the review, you are right the push to feedz.io takes priority so we first run that now.

I also added `continue-on-error: true` to the push to github package repository.

PR #283, opened 2023-03-01
Various cleanups to our github action setup

Mpdreamz:
Few fixes:

- removes docker-run, I think it makes more sense to run on the richer `linux-latest` host.
- Ensure we fetch all tags always through common bootstrap action (the default is depth of `1` which does not get all the tags).
- Ensure dotnet installs using version defined in `global.json` (single source of truth).
- Fix windows build not using single targets (`-s true`) and thus inadvertingly calling `clean` and not publishing the windows test results.
- Remove `.ci` folder completely.


.github/workflows/bootstrap/action.yml
```diff
@@ -0,0 +1,20 @@
...
+name: Bootstrap Checkout
+description: Ensures an action that needs to build has access to git tags
+
+runs:
+  using: "composite"
+  steps:
+    # Ensure we fetch all tags
+    - shell: bash
+      run: |
+        git fetch --prune --unshallow --tags
```
v1v, line 11: I somehow thought the `checkout: depth: 0` would be enough 

```
    # Number of commits to fetch. 0 indicates all history for all branches and tags.
    # Default: 1
    fetch-depth: ''
```

In https://github.com/actions/checkout

v1v (commented):
LGTM, but a minor comment regarding the checkout

PR #306, opened 2023-06-09
Add nullable support to Elastic.CommonSchema.Serilog

Mpdreamz:
Introduce `TryGetScalarString` as extention to `TryGetScalarPropertyValue` that does not assume `ScalarProperty.Value` is never null if returned. 

Further more this enables stricter null checks.


guisantos:
Hi, Is this released already? if so, in which version? Thanks

PR #313, opened 2023-06-23
Ensure Elastic.Extensions.Logging can set auth through config

Mpdreamz:
Previously this was only applied to Cloud connections


sacalata:
Is this currently working on current release?, documentation still says that pass and user is for Cloud only and I haven't been able to get it to work

PR #345, opened 2023-11-07
Use Convert instead of cast for formatting ElapsedMilliseconds

ixnas:
For a project I'm using [SerilogWeb.Classic](https://github.com/serilog-web/classic) to log HTTP requests. The ECS formatter crashed when it tried to cast `ElapsedMilliseconds` from a message template, because `SerilogWeb.Classic` passed a `long`, while the ECS formatter expected a `double`.

Because the property is boxed, the formatter would crash when trying to cast a boxed `long` to a `double`.

I can't find any specification on the specific type that this property should have, so I figured the formatter should be agnostic about the exact numerical type.


Mpdreamz (approved):
Looks good to me! And thank you for catching this pesky bug!

PR #348, opened 2023-12-04
NLog EcsLayout - Added support for ProcessThreadName + MessageTemplate

snakefoot:
"Original"-field was removed from "Log"-section with #194 . Now trying to restore it as "MessageTemplate" property.

Curious where the recommended place to put the original "MesageTemplate". See also: https://discuss.elastic.co/t/log-original-field-lost-with-upgrade-8-6-1-from-1-5-3/348363

Have tried to follow the Elastic Extensions logging, so "MesageTemplate" becomes additional property. See also: https://www.elastic.co/guide/en/ecs-logging/dotnet/current/extensions-logging-data-shipper.html#_example_document

### DIFF
--- a/src/Elastic.CommonSchema.NLog/EcsLayout.cs
+++ b/src/Elastic.CommonSchema.NLog/EcsLayout.cs
@@ -4,7 +4,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.Runtime.Serialization;
 using System.Text;
+using System.Text.Json.Serialization;
+using Elastic.CommonSchema.Serialization;
 using NLog;
 using NLog.Config;
 using NLog.Layouts;
@@ -45,6 +48,8 @@ namespace Elastic.CommonSchema.NLog
 		public EcsLayout()
 		{
 			IncludeEventProperties = true;
+
+			MessageTemplate = "${onhasproperties:${message:raw=true}}";
 
 			LogOriginCallSiteMethod = "${exception:format=method}";
 			LogOriginCallSiteFile = "${exception:format=source}";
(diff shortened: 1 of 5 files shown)


snakefoot:
@Mpdreamz  Really confused by this logic:
```c#
		public static void SetMetaOrLabel(EcsDocument document, string path, object value)
		{
			switch (value)
			{
				case string s:
					document.Labels ??= new Labels();
					document.Labels[path] = s;
					break;
				case bool b:
					document.Labels ??= new Labels();
					document.Labels[path] = b.ToString(CultureInfo.InvariantCulture).ToLowerInvariant();
					break;
				default:
					document.Metadata ??= new MetadataDictionary();
					document.Metadata[path] = value;
					break;
			}
		}
```
Why does is the field-datatype used to control whether it is `MetaData` or `Labels` ?

Notice NLog Layout has support for explict specifying both:
```xml
      <layout xsi:type="EcsLayout">
        <metadata name="MyProperty" layout="MyPropertyValue" /> <!-- repeated, optional -->
        <label name="MyLabel" layout="MyLabelValue" />          <!-- repeated, optional -->
        <tag layout="MyTagValue" />                             <!-- repeated, optional -->
      </layout>
```

Mpdreamz:
@snakefoot this is to align with the ECS guidelines around specifying unknown fields: https://www.elastic.co/guide/en/ecs/current/ecs-custom-fields-in-ecs.html#_the_labels_field

`labels` is explicitly a `Dictionary<string, string>` 
`metadata` is a .NET extension to store unknown complex fields. 

This is a step to move to `labels`. 

I do think `labels` needs to allow complex fields in the ECS spec and be mapped as https://www.elastic.co/guide/en/elasticsearch/reference/current/flattened.html

However in that case in the long run we might run into a separation of `labels` into `numeric_labels` https://github.com/elastic/apm-server/issues/5963.

Will follow up in the ECS repository to discuss long term plans there.

Mpdreamz:
run docs build

Mpdreamz:
run docs-build

Mpdreamz (approved):
LGTM. Thank you for implementing this, @snakefoot.

PR #367, opened 2024-04-02
log4net layout: log latest scope value

andreycha:
**Motivation**
If a property is added to the log scope multiple times, currently all its values are logged in metadata/labels. This is how log4net also behaves but in most of the cases it is undesired behavior. Only latest property value should be included (e.g., Serilog behaves the same way).

Example:
```cs
using (LogicalThreadContext.Stacks["Id"].Push("123"))
{
    log.Info("Log 1"); // currently has "123" in the log metadata

    using (LogicalThreadContext.Stacks["Id"].Push("456"))
    {
        log.Info("Log 2"); // currently has "123 456" in the log metadata; should have "456" instead

        using (LogicalThreadContext.Stacks["Id"].Push("789"))
        {
            log.Info("Log 3"); // currently has "123 456 789" in the log metadata; should have "789" instead
        }
    }
}
```

**Breaking change:** yes (comparing to how log4net behaves by default)

**Changes**
- Log latest property value
- Update to the latest log4net version (2.0.17), this is required to use [`Peek()` method](https://github.com/apache/logging-log4net/pull/101)
- Update test names
- Use file-scoped namespaces
- Update readme to reflect the latest version of generated events

### DIFF
--- a/src/Elastic.CommonSchema.Log4net/Elastic.CommonSchema.Log4net.csproj
+++ b/src/Elastic.CommonSchema.Log4net/Elastic.CommonSchema.Log4net.csproj
@@ -10,6 +10,6 @@
     <ProjectReference Include="..\Elastic.CommonSchema\Elastic.CommonSchema.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="log4net" Version="2.0.15" />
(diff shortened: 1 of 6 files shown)


Mpdreamz:
run docs-build

PR #389, opened 2024-05-28
ci: use updatecli with GitHub secrets

v1v:
Use oblt-actions with GitHub secrets.

No need to use Vault secrets path


Mpdreamz (approved):
LGTM

PR #392, opened 2024-05-28
Add NLog SequenceId to ECS documents

Mpdreamz:


src/Elastic.CommonSchema.NLog/EcsLayout.cs
```diff
@@ -539,6 +539,7 @@ private Event GetEvent(LogEventInfo logEventInfo)
 					? long.Parse(eventSeverity)
 					: GetSysLogSeverity(logEventInfo.Level),
 				Timezone = TimeZoneInfo.Local.StandardName,
+				Sequence = logEventInfo.SequenceID
```
snakefoot, line 542: Notice that SequenceID is an integer. Thus if exceeding more than  2,147,483,647 LogEvents, then you will get bad behavior (imagine a service running for several days). See also: https://github.com/NLog/NLog/issues/4159

snakefoot: Maybe consider using `Layout.FromMethod` just like for TraceId / SpanId.

Mpdreamz: yeah I noticed it was `int` 😢 I assumed NLog had a way of dealing with overflows.

But reading https://github.com/NLog/NLog/issues/4159 SequenceId is actually lazy now and only increments if a layout is requesting it correct?

What happens with folks using `SequenceId` today, most likely overflowing? 

Whats your guidance @snakefoot keep it in so we can move along when NLog fixes this?
Or keep it out till then?

snakefoot: I was thinking to obsolete NLog `LogEventInfo.SequenceID` and instead rely on LayoutRenderer like `${counter}`.

I would probably implement EcsLayout SequenceId like ProcessId has been implemented. By using a Layout, then people can do whatever they want.

snakefoot: Alternative just have a private long-field, and do Interlocked.Increment in the NLog EcsLayout for every LogEvent (Skip the overhead of NLog Layout)

But I don't know how Elastic Sequence are expected to be filled out when, when having single instance restarting frequently (starts counting from Sequence=0), or having multiple process-instances on the same machine.

Mpdreamz: Okay in that case will close this! Thanks @snakefoot

PR #394, opened 2024-05-30
Making logger provider the sole manager of channel lifecycle.

moikmellah:
Fixes a bug where ElasticsearchLoggerProvider passes a reference to EcsDataStreamChannel when instantiating ElasticsearchLogger and disposes of the channel on configuration reload events, leading to ElasticsearchLogger instances hanging on to a disposed channel and silently dropping log events.


moikmellah:
CLA submitted, please let me know if anything else is needed.  Thanks!

PR #396, opened 2024-05-30
Fixing disposed channel after config refresh

moikmellah:
Replacing reference to channel in ElasticsearchLogger with reference to channel provider, to avoid use of disposed channel after config refresh event.


Mpdreamz:
Much needed, thanks for catching this one @moikmellah

Mpdreamz (approved):

PR #399, opened 2024-06-02
#398 set Event Kind to event if not provided

### LINKED ISSUE #398: [FEATURE] Event.Kind should be set in logs
**ECS integration/library project(s) (e.g. Elastic.CommonSchema.Serilog)**:
Elastic.CommonSchema.NLog

**Is your feature request related to a problem? Please describe.**
I want to be able to tell by logging at my logs what kind of event i am looking at. The default should be the most common value which is "[event](https://github.com/elastic/ecs/edit/8.11/docs/fields/field-values.asciidoc)"

**Describe the solution you'd like**
I want the event.kind property to be set to event by default

**Describe alternatives you've considered**
Custom layout

**Additional context**
Note I in some cases will be switching to metric, alert and state once #397 is addressed

thompson-tomo:
With this change the logs will now default to be being recorded as having the event kind of event which is the most common.

Closes #398


src/Elastic.CommonSchema.NLog/EcsLayout.cs
```diff
@@ -534,7 +534,7 @@ private Event GetEvent(LogEventInfo logEventInfo)
 				Action = !string.IsNullOrEmpty(eventAction) ? eventAction : null,
 				Id = EventId?.Render(logEventInfo),
 				Code = eventCode,
-				Kind = !string.IsNullOrEmpty(eventKind) ? eventKind : null,
+				Kind = !string.IsNullOrEmpty(eventKind) ? eventKind : "event",
```
snakefoot, line 537: Please put this logic into the `EventKind`-Layout in your own solution. What is good default for you, might be noise for others.

Ex. one can do this in NLog `eventKind="${event-property:EventKind:whenEmpty=event}"`

src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
```diff
@@ -265,7 +265,7 @@ private static Event GetEvent(LogEvent e)
 				Id = e.TryGetScalarString(SpecialKeys.ActionId, out var actionId)
 					? actionId
 					: null,
-				Kind = e.TryGetScalarString(SpecialKeys.ActionKind, out var actionKind) ? actionKind : null,
+				Kind = e.TryGetScalarString(SpecialKeys.ActionKind, out var actionKind) ? actionKind : "event",
```
snakefoot, line 268: You can add a Serilog-enricher that always inserts the ActionKind-property in your own solution. What is good default for you, might be noise for others.

src/Elastic.CommonSchema.NLog/EcsLayout.cs
thompson-tomo: My understanding is that on reading the docs, event kind should be set in the logs. However if the noise is a genuine concern, perhaps we include multiple layouts in the package 1 for each kind.

Mpdreamz:
Hi @thompson-tomo thanks for the PR.

The ecs categorization fields should be empty by default as per: https://www.elastic.co/guide/en/ecs/current/ecs-category-field-values-reference.html


```
If your events don’t match any of these categorization values, you should leave the fields empty. This will ensure you can start populating the fields once the appropriate categorization values are published, in a later release.
```

If someone wants event.kind to be `event` by default I think they have to actively choose to do so themselves.

PR #439, opened 2024-09-16
Introduce Console formatter for Microsoft.Extensions.Logging.Console"

Mpdreamz:
- Introduce Console formatter Microsoft.Extensions.Logging
- Add README contents

### DIFF
--- a/src/Elastic.Extensions.Logging.Common/LogEventBuilderExtensions.cs
+++ b/src/Elastic.Extensions.Logging.Common/LogEventBuilderExtensions.cs
@@ -1,0 +1,212 @@
+using System.Collections;
+using System.Text;
+using Elastic.CommonSchema;
+using Microsoft.Extensions.Logging;
+
+namespace Elastic.Extensions.Logging.Common;
+
+/// <summary> </summary>
+public static class LogEventBuilderExtensions
+{
+	/// <summary> </summary>
+	public static void AddScopeValues(this LogEvent logEvent, IExternalScopeProvider? scopeProvider, ILogEventCreationOptions options)
+	{
+		void AddScopeValue<TLocalState>(TLocalState scope, LogEvent log)
+		{
+			if (scope is null) return;
+
+			log.Labels ??= new Labels();
+			log.Scopes ??= new List<string>();
+
+
+			var scopeValues = (scope as IEnumerable<KeyValuePair<string, object>>)?.ToList();
+			var scopeName = scopeValues != null && scopeValues.Any(kv => kv.Key == "{OriginalFormat}")
+				? scope.ToString()
+				: FormatValue(scope, options, 0, scope.GetType().Name);
+			log.Scopes.Add(scopeName);
+
+			if (scopeValues == null) return;
+
+			foreach (var kvp in scopeValues)
+				AssignStateOrScopeLabels(logEvent, kvp, options);
+		}
+
+		scopeProvider?.ForEachScope((o, @event) => AddScopeValue(o, @event), logEvent);
+	}
+
+	/// <summary> </summary>
+	public static void AddStateValues<TState>(this LogEvent logEvent, TState state, ILogEventCreationOptions options)
+	{
+		if (state is not IEnumerable<KeyValuePair<string, object>> stateValues) return;
+
+		foreach (var kvp in stateValues)
+			AssignStateOrScopeLabels(logEvent, kvp, options);
+	}
+
+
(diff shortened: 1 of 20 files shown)


src/Elastic.Extensions.Logging.Common/LogEventBuilderExtensions.cs
```diff
@@ -0,0 +1,212 @@
...
+	{
+		void AddScopeValue<TLocalState>(TLocalState scope, LogEvent log)
+		{
+			if (scope is null) return;
+
+			log.Labels ??= new Labels();
+			log.Scopes ??= new List<string>();
+
+
+			var scopeValues = (scope as IEnumerable<KeyValuePair<string, object>>)?.ToList();
```
stevejgordon, line 22: I think this is fine for now, but maybe we should consider a lower allocation approach with a rented array for these or even `stackalloc` for low volumes?

src/Elastic.Extensions.Logging.Console/EcsConsoleFormatterOptions.cs
```diff
@@ -0,0 +1,28 @@
...
+	public bool IncludeProcess { get; set; } = true;
+
+	/// <inheritdoc cref="IEcsDocumentCreationOptions.IncludeUser"/>
+	public bool IncludeUser { get; set; } = true;
+
+	/// <inheritdoc cref="IEcsDocumentCreationOptions.IncludeActivityData"/>
+	public bool IncludeActivityData { get; set; } = true;
+
+	/// <inheritdoc cref="ILogEventCreationOptions.Tags"/>
+	public string[]? Tags { get; set; }
```
stevejgordon, line 24: I'm becoming a fan of avoiding nulls like this. Would making this default to an empty array and making it non-nullable be an option?

src/Elastic.Extensions.Logging.Console/README.md
```diff
@@ -0,0 +1,24 @@
+# Elastic.Extensions.Logging.Console
+
+This package includes formatters and extension methods for `Microsoft.Extensions.Logging.Console` to make it easy to write ECS formatted logs to consoleoutput.
+
+May be used with `Elastic.Extensions.Logging` to write ECS documents directly to Elasticsearch / Elastic Cloud. 
+
+
+## Usage
+
+The console logging provider and formatter can be setup using a simple extension method.
```
stevejgordon, line 10:
```suggestion
The console logging provider and formatter can be set up using a simple extension method.
```

src/Elastic.Extensions.Logging.Common/LogEventBuilderExtensions.cs
```diff
@@ -0,0 +1,212 @@
...
+				logEvent.Http ??= new Http();
+				logEvent.Http.RequestId = requestId;
+				return true;
+			case "RequestPath" when value is string path:
+				logEvent.Url ??= new Url();
+				logEvent.Url.Path = path;
+				return true;
+			// ReSharper disable once UnusedVariable
+			case "Protocol" when value is string protocol:
+				// TODO protocol
```
stevejgordon, line 74: Was this missed or for follow up?

stevejgordon (commented):
A few comments to consider, nothing blocking if we want to get this in and follow up. 

There's also use of `StringBuilder`. Given this is high-volume, do we want to consider refactoring to rented array/`Span` based processing to improve performance?

src/Elastic.Extensions.Logging.Common/LogEventBuilderExtensions.cs
Mpdreamz: Ahh this was moved pasted from elsewhere into the `Common` lib. 

I think we have no place to put it currently so to the TODO is not actionable

src/Elastic.Extensions.Logging.Common/LogEventBuilderExtensions.cs
Mpdreamz: I think we should but this is already in place today, only lights up in the PR because we moved it to a new shared library. 

Shall we follow up with that in another PR?

src/Elastic.Extensions.Logging.Console/EcsConsoleFormatterOptions.cs
Mpdreamz: It would but this would be breaking since we inherit this from `ILogEventCreationOptions` and `ElasticsearchLoggerOptions` who also implements it is already released.

stevejgordon: Gotcha! Fair enough.

stevejgordon (approved):
LGTM

Issue #411, opened 2024-07-10
[FEATURE] Elastic.Serilog.Sinks - Allow flushing messages when Serilog sink is disposed


kudmax:
**ECS integration/library project(s)**:
`Elastic.Serilog.Sinks`

**Is your feature request related to a problem? Please describe.**
If the application shuts down without suitable timeout, all logged messages waiting in the internal buffer will be lost. Unfortunately, there are no flushing capabilities available to process the internal buffer before its `OutboundBufferMaxLifetime` expires or the maximum number of elements is reached.

**Describe the solution you'd like**
The `Serilog` static API includes a `CloseAndFlush()` method that finalizes the underlying configured logger. When invoked, this method disposes the logger, which in turn disposes of all its sinks. If the sink needs to complete logging in a specific way (such as flushing messages temporarily residing in the internal buffer), it should implement `IDisposable` or `IAsyncDisposable`.

Considering this, I would propose supplying the Elasticsearch sink with IDisposable implementation that would ensure proper flushing in the hosted channel at dispose, preventing the loss of buffered messages.

lowern1ght:
+1

Mpdreamz:
Thanks for highlighting this omission! This will be covered once https://github.com/elastic/ecs-dotnet/pull/440 lands

PR #440, opened 2024-09-16
Update to latest ingest that attempts to flush out buffer before fully disposing

Mpdreamz:
This also updates our serology sink to implement `IDisposable` so it waits to flush as well. See also #411


stevejgordon (approved):
LGTM

PR #453, opened 2024-10-03
Remove WriteEvent usages

Mpdreamz:
- Ensure correct embdedded logical name for UParser.regexes.cs
- Remove WriteEvent usages


stevejgordon (approved):
LGTM

PR #454, opened 2024-10-03
Test windows action should call tests explictly now that release target no longer depends on test

Mpdreamz:


stevejgordon (approved):
LGTM

PR #469, opened 2024-12-10
Updated to NET8 since NET6 is obsolete

snakefoot:


snakefoot:
@Mpdreamz Possible to re-trigger the build one more time?

Mpdreamz (approved):
Excellent work as always @snakefoot ! Sorry it took a while on my end to review and get this in.

.github/workflows/bootstrap/action.yml
```diff
@@ -14,7 +14,9 @@ runs:
     # Install .NET version as mandated by global.json
     - uses: actions/setup-dotnet@v4
       with:
-          global-json-file: global.json
```
flobernd, line 17: What's the reason for this change? Do we need both, net6 and net8 installed?

flobernd (approved):
LGTM! Just one Q.

.github/workflows/bootstrap/action.yml
Mpdreamz: This was needed for `nupkg-validator` but the bump to `0.5.0` should have fixed this 🤞 will revert this to verify. 

Starting with .NET 9.0 the SDK check on dotnet tools is relaxed which is great.

flobernd (approved):

Mpdreamz:
run docs-build

PR #495, opened 2025-04-20
ci: group dependabot github action bump

kruskall:
reduce pull request noise for github action bump

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,6 @@ updates:
   - package-ecosystem: "github-actions"
     directories:
       - "/"
-      - "/.github/actions/*"
       - "/.github/workflows/*"
     reviewers:
       - "elastic/observablt-ci"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -3,7 +3,9 @@ updates:
 
   # GitHub actions
   - package-ecosystem: "github-actions"
-    directory: "/"
+    directories:
+      - "/"
+      - "/.github/actions/*"
```
v1v, line 8:
```suggestion
      - "/.github/actions/*"
      - "/.github/workflows/*"
```

v1v (approved):

.github/dependabot.yml
v1v: > as well as the action.yml/action.yaml file from the root directory.

IIUC, it will not consider nested folders, as we have in our case:

- `.github/workflows/bootstrap/action.yml`

Maybe I'm wrong, but, I recall we had to use the folder to make this working.

v1v (approved):

.github/dependabot.yml
v1v: Thanks for double checking it 🙇

v1v (approved):

Mpdreamz (approved):

PR #514, opened 2025-06-11
Ensure we can handle more data types in ElapsedMillisecond/Elapsed from serilog

Mpdreamz:


src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
```diff
@@ -300,6 +306,19 @@ private static Event GetEvent(LogEvent e)
...
+
+			double? ToDouble()
+			{
+				try
+				{
+					return Convert.ToDouble(elapsedMsObj.Value);
+				}
+				catch
+				{
+					// ignored
```
reakaleek, line 318: Q: Would it make sense to put a warning here? I feel like returning `null` could have unexpected behaviour for the user? Asking this without knowing all the context.

reakaleek (approved):

src/Elastic.CommonSchema.Serilog/LogEventConverter.cs
Mpdreamz: Normally yeah, but since this is after we exhausted all the regular casts its best effort from here on out.

flobernd (approved):

JeremyBessonElastic (approved):
LGTM

Minor comment: the logic with all the ?/if all imbricated might deserve a simpler form with only if and a private method

PR #523, opened 2025-06-27
fix/test fallback serilog

Mpdreamz:
- Test fallback
- update
- Ensure fallback mechanism works with the latest ingest-dotnet


JeremyBessonElastic (approved):
LGTM

